PR #5, opened 2020-10-08
Code cleanup and add golangci-lint

darkowlzz:
- Move `kubebuilder:scaffold:builder` marker & waitgroup add to proper locations.
- Remove api/ dir - no CRD APIs.
- Add golangci-lint config.
- Cleanup code based on golangci-lint errors.
- Change `ResultMetric.Increment()` signature from `metrics` package to avoid the need to check unrelated passed in error.
- Add new workflow job to run golangci-lint.


croomes (approved):
Looks great - thanks!

PR #7, opened 2020-10-20
Use PV name as service name

croomes:
The RWX Service should be named afater the PV, not the PVC that created it.  The PV name is unique across the cluster, and allocated immediately before the Service is created.  The PVC name is not suitable as it is common for applications to name the PVC the same as the application Service.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/go-logr/logr v0.1.0
 	github.com/golang/protobuf v1.4.2 // indirect
+	github.com/google/uuid v1.1.1
 	github.com/hashicorp/go-multierror v1.1.0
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/go-logr/logr v0.1.0
 	github.com/golang/protobuf v1.4.2 // indirect
+	github.com/google/uuid v1.1.1
```
darkowlzz, line 8: We also need vendored deps update. `make tidy`.

darkowlzz (approved):
LGTM!
Thanks.

go.mod
darkowlzz: It already exists as a transitive dependency https://github.com/storageos/api-manager/tree/master/vendor/github.com/google/uuid .

PR #13, opened 2020-12-03
Add node delete controller

croomes:
Adds a controller to watch for Kubernetes Node delete events, and to remove the node from the StorageOS cluster if safe.

### DIFF
--- a/controllers/node-delete/controller.go
+++ b/controllers/node-delete/controller.go
@@ -1,0 +1,121 @@
+package nodedelete
+
+import (
+	"encoding/json"
+
+	"github.com/go-logr/logr"
+	"github.com/storageos/api-manager/internal/pkg/storageos"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	// DriverName is the name of the StorageOS CSI driver.
+	DriverName = "csi.storageos.com"
+
+	// DriverAnnotationKey is the annotation that stores the registered CSI
+	// drivers.
+	DriverAnnotationKey = "csi.volume.kubernetes.io/nodeid"
+)
+
+// Reconciler reconciles a Node object by deleting the StorageOS node object
+// when the corresponding Kubernetes node is deleted.
+type Reconciler struct {
+	client.Client
+	Log logr.Logger
+	api storageos.NodeDeleter
+}
+
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
+
+// NewReconciler returns a new Node delete reconciler.
+func NewReconciler(api storageos.NodeDeleter, k8s client.Client) *Reconciler {
+	return &Reconciler{
+		Client: k8s,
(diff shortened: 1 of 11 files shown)


controllers/node-delete/controller.go
```diff
@@ -0,0 +1,88 @@
+package nodedelete
+
+import (
+	objectv1 "github.com/croomes/object-reconciler/controller/v1"
```
croomes, line 4: I'll move this package to storageos GH namespace.

croomes: done by removing the dependency, it wasn't adding much.

controllers/node-delete/controller.go
```diff
@@ -0,0 +1,91 @@
...
+	err := r.api.DeleteNode(req.Name)
+	if err != nil && err != storageos.ErrNamespaceNotFound {
+		// Re-queue without error.  We will get frequent transient errors, such
+		// as version conflicts or locked objects - that's ok.  Even refusal to
+		// delete due to the node still showing as online should be seen as
+		// transient and should not raise an error.
+		return ctrl.Result{Requeue: true}, nil
+	}
+	// We can't return an event as it requires an object, and we can't rely on
+	// the object still being present.
```
darkowlzz, line 63: Maybe this comment can be removed. We don't publish any event in this controller.

controllers/node-delete/controller.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Update determines whether an object update should trigger a reconcile.
+func (ChangePredicate) Update(event.UpdateEvent) bool {
+	return false
+}
+
+// Delete determines whether an object delete should trigger a reconcile.
+func (ChangePredicate) Delete(e event.DeleteEvent) bool {
+	// Ignore objects without metadata.
+	return e.Meta != nil
```
darkowlzz, line 85: We should also be able to check the node annotations to find out if storageos is installed on the node and ignore the events from nodes that don't have storageos.

controllers/node-delete/controller.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Reconcile deletes the StorageOS node.  The delete will fail if the control
+// plane has not yet determined that the node is offline.  Any errors will
+// result in a requeue, with standard back-off retries.
+//
+// Events are not sent as they require an object. By this point the namespace
+// object will no longer be available.
+func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	err := r.api.DeleteNode(req.Name)
+	if err != nil && err != storageos.ErrNamespaceNotFound {
```
darkowlzz, line 55: Should this be node not found error?
I don't think nodes are namespaced.

controllers/node-delete/controller.go
```diff
@@ -0,0 +1,116 @@
...
+// driver annotation.  The annotation is added by the drivers' node registrar,
+// so should only be present on nodes that have run StorageOS.
+func hasStorageOSDriverAnnotation(annotations map[string]string) bool {
+	drivers, ok := annotations[DriverAnnotationKey]
+	if !ok {
+		return false
+	}
+	driversMap := map[string]string{}
+	if err := json.Unmarshal([]byte(drivers), &driversMap); err != nil {
+		return false
```
darkowlzz, line 110: Logging this error would be helpful in debugging when it fails. Avoiding silent failures.

darkowlzz (approved):
LGTM!
A lot of nice tests. Thanks.

controllers/node-delete/controller.go
croomes: Agreed, added

PR #15, opened 2020-12-18
StorageOS API interface and mocks

croomes:
This makes it easier to unit test code in the StorageOS API wrapper (internal/pkg/storageos) by accepting an interface for the client and auto-generating mocks.

This is a pre-requisite for node label sync where there will be some logic in the wrapper that needs to be tested.


darkowlzz (approved):
LGTM!
Thanks for the generated mocks. Very nice.

PR #16, opened 2021-01-06
Update dependencies

croomes:
Updates to latest controller-runtime (v0.7.0) required for kubebuilder v3 and operator-toolkit.  Also pulls in updates to logger, client-go, etc.

Adds missing perms for namespace delete controller.


darkowlzz (approved):
Everything looks good to me.
⚠️ Since we vendor the dependencies, we should also update the vendor dir content.
With this, everything will be up-to-date and it'll be very easy to convert the project to kubebuilder v3.
Thanks.

PR #17, opened 2021-01-07
Add action to ensure modules tidy & vendored

croomes:
Ensures changes from `go mod tidy` and `go mod vendor` have been commited.

Failure when dependency missing from vendor: https://github.com/storageos/api-manager/runs/1662859506?check_suite_focus=true

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           go-version: '1.15.2'
       - name: Run go tidy
-        run: make tidy && git diff --exit-code && test -z "$(git ls-files --exclude-standard --others)"
+        run: make tidy && git diff --exit-code && test -z "$(git ls-files --exclude-standard --others | tee /dev/fd/2)"
       - name: Run go test
         run: make test
       - name: Run go build
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -16,14 +16,6 @@ jobs:
         with:
           args: --timeout=5m
 
-  tidy:
-    name: Tidy
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


croomes:
I think the `lint` test waiting for status to be reported is because I renamed it to be consistent.  It should be ok to ignore.

.github/workflows/test-build.yml
```diff
@@ -7,15 +7,23 @@ env:
...
         with:
           args: --timeout=5m
 
+  tidy:
+    name: Tidy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run go tidy
+        run: make tidy && git diff --exit-code && test -z "$(git ls-files --exclude-standard --others)"
```
darkowlzz, line 25: Maybe tidy should be a part of the following `test-build` job. Tidy runs go mod. The version of go should be the same when running tidy, test and build to have consistent results.

darkowlzz: Also, it'd be helpful to print the diff at the end when it fails.

croomes: Yes, that's better.  I've moved the job, and it now outputs dirty files to stderr.  It's still not super-obvious when they're just new files but for changes you get full diffs.  A missing vendor module shows up as in https://github.com/storageos/api-manager/runs/1664181588?check_suite_focus=true#step:4:567

The last lines are the missing files:
```
vendor/gomodules.xyz/jsonpatch/v2/LICENSE
vendor/gomodules.xyz/jsonpatch/v2/go.mod
vendor/gomodules.xyz/jsonpatch/v2/go.sum
vendor/gomodules.xyz/jsonpatch/v2/jsonpatch.go
```

darkowlzz (approved):
Excluding standard files and including untracked files in the output will be very helpful.
Thanks.
LGTM!

PR #18, opened 2021-01-07
Build and tests should use vendored deps

croomes:
Change build and tests to use the vendored dependencies.

Also removes the code in the Dockerfile to selectively copy source files since it was intended to speed up the build by not copying in vendored deps (which we now want to do).


darkowlzz (approved):
LGTM!
Thanks.

PR #19, opened 2021-01-07
Migrate node delete to operator toolkit

croomes:
This converts the node delete controller to use the external-object-sync helper from the new operator toolkit.

It adds a configurable periodic garbage collection, to ensure any nodes that we removed while the api-manager wasn't running will be eventually removed from StorageOS.  The usual API protections are still present - the node must be offline and not host any master deployments before it will be deleted.


controllers/node_delete_test.go
```diff
@@ -165,4 +168,14 @@ var _ = Describe("Node Delete controller", func() {
 		})
 	})
 
+	Context("When starting after ak8s Node has been deleted but is still in StorageOS", func() {
```
darkowlzz, line 171:
```suggestion
	Context("When starting after a k8s Node has been deleted but is still in StorageOS", func() {
```

darkowlzz (approved):
LGTM!
Nice tests and `IgnoreFuncs` predicate.

PR #20, opened 2021-01-08
Migrate namespace delete to operator toolkit

croomes:
Updates the namespace delete controller to use the operator toolkit, and adds garbage collection every 1h by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,14 +91,16 @@ The following flags are supported:
     	Maximum time to wait for new Kubernetes objects to appear. (default 20s)
   -kubeconfig string
     	Paths to a kubeconfig. Only required if out-of-cluster.
-  -master --kubeconfig
-    	(Deprecated: switch to --kubeconfig) The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.
   -metrics-addr string
     	The address the metric endpoint binds to. (default ":8080")
+  -namespace-delete-gc-interval duration
+    	Frequency of node garbage collection. (default 1h0m0s)
   -namespace-delete-workers int
     	Maximum concurrent namespace delete operations. (default 5)
+  -node-delete-gc-interval duration
+    	Frequency of node garbage collection. (default 1h0m0s)
   -node-delete-workers int
-    	Maximum concurrent node delete operations. (default 5)      
+    	Maximum concurrent node delete operations. (default 5)   
 ```
 
 ## Setup/Development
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -91,14 +91,16 @@ The following flags are supported:
     	Maximum time to wait for new Kubernetes objects to appear. (default 20s)
   -kubeconfig string
     	Paths to a kubeconfig. Only required if out-of-cluster.
-  -master --kubeconfig
-    	(Deprecated: switch to --kubeconfig) The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.
   -metrics-addr string
     	The address the metric endpoint binds to. (default ":8080")
+  -namespace-delete-gc-interval duration
+    	Frequency of node garbage collection. (default 1h0m0s)
```
darkowlzz, line 97:
```suggestion
    	Frequency of namespace garbage collection. (default 1h0m0s)
```

controllers/namespace-delete/controller.go
```diff
@@ -3,83 +3,51 @@ package nsdelete
...
-		// as version conflicts or locked objects - that's ok.  Even refusal to
-		// delete due to remaining volumes should be seen as transient and
-		// should not raise an error.
-		return ctrl.Result{Requeue: true}, nil
+// Delete receives a k8s object that's been deleted and calls the StorageOS api
+// to remove it from management.
+func (c Controller) Delete(ctx context.Context, obj client.Object) error {
+	err := c.api.DeleteNamespace(obj.GetName())
+	if err != nil && err != storageos.ErrNodeNotFound {
+		return errors.Wrap(err, "re-queing operation")
```
darkowlzz, line 41: maybe `requeuing` ?

darkowlzz (approved):
LGTM!

controllers/namespace-delete/controller.go
avestuk: Should that be checking for `storageos.ErrNamespaceNotFound` or similar?

README.md
croomes: done

controllers/namespace-delete/controller.go
croomes: Very well spotted!  This wasn't caught by the test for 2 reasons:
1. envtest doesn't propagate namespace delete events, so the test is skipped when running in envtest, which is all that is available in the pipelines.
2. I hadn't run it manually against a real cluster (`USE_EXISTING_CLUSTER=true go test -v github.com/storageos/api-manager/controllers`), and when I did it passed when it should have failed.  

I've now updated the test (also for node delete), confirmed it fails with the copypasta mistake, and then fixed the mistake.

PR #21, opened 2021-01-12
Update to new external object sync

croomes:
Updates to the latest version of the operator toolkit which will be needed for label sync.  It changes the external object sync, see example: https://github.com/darkowlzz/operator-toolkit/blob/14649a309dc6585984e6f464788e4853dda6bacb/example/controllers/external-game-sync_controller.go


darkowlzz (approved):
LGTM!

controllers/namespace-delete/controller.go
```diff
@@ -6,7 +6,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	objectv1 "github.com/darkowlzz/operator-toolkit/controller/external-object-sync/v1"
+	syncv1 "github.com/darkowlzz/operator-toolkit/controller/sync/v1"
```
darkowlzz, line 9: This and other imports are not separated correctly for some reason.
I think we only separate native and external imports.

controllers/node-delete/controller.go
```diff
@@ -6,7 +6,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	objectv1 "github.com/darkowlzz/operator-toolkit/controller/external-object-sync/v1"
+	syncv1 "github.com/darkowlzz/operator-toolkit/controller/sync/v1"
```
darkowlzz, line 9: Same weird import separation like in namespace-delete controller.

controllers/namespace-delete/controller.go
croomes: Yes, will fix - thanks

controllers/node-delete/controller.go
croomes: Yes, will fix - thanks

PR #22, opened 2021-01-12
Node label sync

croomes:
Adds node label sync with a periodic resync.

Reserved labels such as `storageos.com/computeony` get applied via specific API calls and the remainder get set as a blob in an UpdateNode() API call.

Labels are also synced when the StoragOS CSI annotation is added, ensuring that the compute-only label is applied as quickly as possible after the node is available in StorageOS.


controllers/node-label/README.md
```diff
@@ -0,0 +1,51 @@
...
+by setting the `storageos.com/computeonly=true` label on a Kubernetes node.
+This controller ensures that the behaviour is applied to the StorageOS node.
+
+## Trigger
+
+The controller reconcile will trigger on any Kubernetes Node label update event
+where the node has the StorageOS CSI driver annotation.
+
+The annotation is added by the CSI node driver registrar when StorageOS starts
+on the node.  Once added, it is not removed.
```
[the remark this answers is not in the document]
croomes: No, this is something that we could add, but at the moment the annotation is never removed, even if StorageOS is uninstalled on the node.  There isn't an easy way for us to determine if StorageOS has been uninstalled or just restarting.  Perhaps by hooking into the DeleteNode API call?

internal/pkg/storageos/node_labels.go
```diff
@@ -0,0 +1,174 @@
...
+	return errors.ErrorOrNil()
+}
+
+// EnsureUnreservedNodeLabels applies a set of labels to the StorageOS node if different.
+// Existing labels will be overwritten.  The set of labels must not include
+// StorageOS reserved labels.
+func (c *Client) EnsureUnreservedNodeLabels(name string, labels map[string]string) error {
+	funcName := "ensure_unreserved_node_labels"
+	start := time.Now()
+	defer func() {
```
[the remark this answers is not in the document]
croomes: Unless it's called in a function, then the params to Observe will be set at the time the defer is defined, rather than when it is called.

internal/pkg/storageos/node_labels.go
```diff
@@ -0,0 +1,174 @@
...
+	if _, _, err = c.api.UpdateNode(ctx, node.Id, api.UpdateNodeData{Labels: labels, Version: node.Version}); err != nil {
+		return observeErr(err)
+	}
+	return observeErr(nil)
+}
+
+// EnsureComputeOnly ensures that the compute-only behaviour has been applied to
+// the StorageOS node.
+func (c *Client) EnsureComputeOnly(name string, enabled bool) error {
+	funcName := "ensure_compute_only"
```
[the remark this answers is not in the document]
croomes: yeah, I might forget to update the func name string at some point but would prefer to keep it simple/readable!

internal/pkg/storageos/node.go
```diff
@@ -27,12 +27,40 @@ var (
...
+	nodeLabels := make(map[string]Object)
+	for _, node := range nodes {
+		nodeLabels[node.GetName()] = node
+	}
+
+	return nodeLabels, nil
+}
+
+// NodeNamespacedNames returns a list of all StorageOS node objects.
+func (c *Client) NodeNamespacedNames() ([]types.NamespacedName, error) {
```
avestuk, line 62: I'm not really sure I understand the naming of this function. At first glance it seems quite confusing because nodes are not namespaced. 

In the Namespace delete controller you are using `ListNamespaces()` and returning the same `[]types.NamespacedName` so perhaps just calling the function `ListNodes()` would be easier to understand and more consistent?

controllers/node-label/controller.go
```diff
@@ -0,0 +1,78 @@
...
+// StorageOS reserved labels are validated and applied first, then the remaining
+// unreserved labels are applied.
+//
+// Any errors will result in a requeue, with standard back-off retries.
+//
+// There is no label sync from StorageOS to Kubernetes.  This is intentional to
+// ensure a simple flow of desired state set by users in Kubernetes to actual
+// state set on the StorageOS node.
+func (c Controller) Ensure(ctx context.Context, obj client.Object) error {
+	if err := c.api.EnsureNodeLabels(obj.GetName(), obj.GetLabels()); err != nil {
```
darkowlzz, line 40: Can we update the `EnsureNodeLabels()` function signature to accept the context and pass it forward?
I see `EnsureUnreservedNodeLabels()` and `EnsureComputeOnly()` creating their own context.

darkowlzz (commented):
Looks very good to me.
Nice tests using generated go mocks.
Left a suggestion about passing the context from controller to the storageos client.

internal/pkg/storageos/node_labels.go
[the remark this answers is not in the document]
croomes: Discussed with @darkowlzz - we'll leave as-is for now.

internal/pkg/storageos/node.go
croomes: After discussing, I'm going to refactor `ListNamespaces()` to behave more like this.  I'll do it in a separate PR.

controllers/node-label/controller.go
croomes: As discussed, I've done just for the functions used in this PR.  I'll open a separate PR to apply the function signature consistently across the rest.

internal/pkg/storageos/node.go
```diff
@@ -43,10 +43,9 @@ func (c *Client) NodeObjects() (map[string]Object, error) {
 		return e
 	}
 
-	ctx, cancel := context.WithTimeout(c.ctx, DefaultRequestTimeout)
-	defer cancel()
+	ctx = c.AddToken(ctx)
 
-	nodes, _, err := c.api.ListNodes(ctx)
+	nodes, _, err := c.api.ListNodes(c.AddToken(ctx))
```
darkowlzz, line 48: The token is added above. Just passing the context should be enough.

croomes: Well spotted - fixed

darkowlzz (approved):
LGTM!

PR #25, opened 2021-01-20
Use Object interface

croomes:
Applies the Object interface changes to node-delete and namespace-delete controllers.


darkowlzz (approved):
Nice usage of the Object interface.
LGTM!

PR #27, opened 2021-01-28
Run resync on startup after a configurable delay

croomes:
Adds flags to configure how long each controller should wait before doing a full resync/garbage collection.  These are set low by default so that they run almost immediately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,18 +93,26 @@ The following flags are supported:
     	Paths to a kubeconfig. Only required if out-of-cluster.
   -metrics-addr string
     	The address the metric endpoint binds to. (default ":8080")
+  -namespace-delete-gc-delay duration
+    	Startup delay of initial namespace garbage collection. (default 20s)
   -namespace-delete-gc-interval duration
     	Frequency of namespace garbage collection. (default 1h0m0s)
   -namespace-delete-workers int
     	Maximum concurrent namespace delete operations. (default 5)
+  -node-delete-gc-delay duration
+    	Startup delay of initial node garbage collection. (default 30s)
   -node-delete-gc-interval duration
     	Frequency of node garbage collection. (default 1h0m0s)
   -node-delete-workers int
     	Maximum concurrent node delete operations. (default 5)
+  -node-label-resync-delay duration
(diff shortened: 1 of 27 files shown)


avestuk (approved):

README.md
```diff
@@ -93,18 +93,26 @@ The following flags are supported:
...
   -node-delete-workers int
     	Maximum concurrent node delete operations. (default 5)
+  -node-label-resync-delay duration
+    	Startup delay of initial node label resync. (default 10s)
   -node-label-resync-interval duration
     	Frequency of node label resync. (default 1h0m0s)
   -node-label-sync-workers int
     	Maximum concurrent node label sync operations. (default 5)
+  -pvc-label-resync-delay duration
+    	Startup delay of initial PVC label resync. (default 1s)
```
darkowlzz, line 115: In my testing, I've only seen failures with zero delay, the controller components aren't available immediately. But maybe 1s is also risky. Just to be safe, 5s should be better.
The default 10s in the core sync package is more than enough time. It's intended to be safe always.

controllers/suite_test.go
```diff
@@ -25,6 +25,17 @@ const (
 	// ready to process events.  The manager starts asychronously, so without a
 	// wait, events may be missed.
 	managerWaitDuration = 1 * time.Second
+
+	// defaultWorkers is the number of concurrent requests that test controllers
+	// can operate on in paralllel.
+	defaultWorkers = 1
+
+	// defaultSyncDelay sets the sync controller to run immediately.
+	defaultSyncDelay = time.Duration(0)
```
darkowlzz, line 34: It would be interesting to see if this works fine even when running envtest in an existing cluster.
We may have to change this zero delay in the future or if we start observing flaky tests.

darkowlzz (approved):
The changes LGTM!
Left an inline suggestion about increasing the delay a little to be safe.

README.md
croomes: Good idea - increased to 5s

controllers/suite_test.go
croomes: Increased to 1s, will keep an eye on it.

PR #29, opened 2021-01-29
Include reserved labels in node label update

croomes:
Similar to volumes, the control plane API expects the current reserved labels to be included in the map of labels sent during a node update, and they must have the same value as there are separate API endpoints to update them.

### DIFF
--- a/internal/pkg/storageos/node_labels_test.go
+++ b/internal/pkg/storageos/node_labels_test.go
@@ -260,3 +260,164 @@ func TestClient_EnsureNodeLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestClient_EnsureUnreservedNodeLabels(t *testing.T) {
+	tests := []struct {
+		name    string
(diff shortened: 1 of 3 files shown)


darkowlzz (approved):
Nice usage of mock expectations.
LGTM!

internal/pkg/storageos/node_labels_test.go
```diff
@@ -260,3 +260,165 @@ func TestClient_EnsureNodeLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestClient_EnsureUnreservedNodeLabels(t *testing.T) {
+	tests := []struct {
+		name    string
+		labels  map[string]string
+		prepare func(name string, m *mocks.MockControlPlane)
+		wantErr bool
```
darkowlzz, line 269: Maybe remove this? Unused.

croomes: removed

PR #30, opened 2021-02-01
Reset api after token refresh error

croomes:
If the periodic token refresh fails, then force an API client reset, which re-creates the client after re-reading the auth secret.  This ensures that an old expired token isn't retried on the next refresh interval.


darkowlzz (approved):
Nice fix.
LGTM!

PR #31, opened 2021-02-01
Standardise on ObjectKeys for all API function signatures

croomes:
No functional changes - this applies the same function signature that was used for Volume API calls to nodes and namespaces, and consolidates some similar functions.

The shared volume `SetExternalEndpoint()` was left as-is as the volume ID is known and used instead as an optimisation.


darkowlzz (approved):
LGTM!
Thanks.

PR #32, opened 2021-02-02
Add OpenTelemetry tracing with Jaeger backend

croomes:
Adds tracing capability to the controllers, sending data to Jaeger.

The object controllers are pretty straightforward but it's a little more awkward to add to the SharedVolume controller without refactoring (which I'd like to avoid atm).


darkowlzz (approved):
LGTM!

PR #33, opened 2021-02-11
Run shared volume controller with manager

croomes:
Code simplification to let controller manager handle startup and shutdown of non-k8s controllers like the SharedVolume controller.


darkowlzz (approved):
LGTM!
Thanks.

PR #34, opened 2021-02-17
Add fencing controller

croomes:
Adds pod fast failover, comparable to the v1 feature.  Pods must have the `storageos.com/fenced=true` label set to be considered for fencing.

The implementation specifics are described in https://github.com/storageos/api-manager/blob/fencing/controllers/fencer/README.md

Includes full tracing support.


controllers/fencer/fencer.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// Process each pod independently.
+	for _, pod := range podList.Items {
+		pod := pod
+		log := r.log.WithValues("name", pod.GetName(), "namespace", pod.GetNamespace())
+		log.V(4).Info("evaluating pod for fencing")
+
+		// Ignore pods that don't have the fenced label set.
+		fenced := false
+		for k, v := range pod.GetLabels() {
```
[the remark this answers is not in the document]
croomes: Yes, it makes a lot more sense - thanks!

controllers/fencer/fencer.go
```diff
@@ -0,0 +1,292 @@
...
+	for _, pvc := range pvcs {
+		if pvc.Spec.VolumeName != "" {
+			volKeys = append(volKeys, types.NamespacedName{Name: pvc.Spec.VolumeName, Namespace: pvc.Namespace})
+		}
+	}
+
+	// Skip pod if no StorageOS PVCs or PVs.
+	if len(volKeys) == 0 {
+		span.SetStatus(codes.Ok, "no storageos pvcs")
+		log.Info("skipping pod with no storageos pvcs")
```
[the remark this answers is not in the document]
croomes: In this case the pod name is added as a value to the logger above, so all subsequent logs will have it included:
```
log := r.log.WithValues("pod", pod.GetName(), "namespace", pod.GetNamespace())
```

controllers/fencer/fencer.go
```diff
@@ -0,0 +1,292 @@
...
+	for _, volume := range volumes {
+		if !volume.IsHealthy() {
+			log.Info("fencer skipping pod with unhealthy volume", "volume", volume.GetName())
+			canFailover = false
+			break
+		}
+	}
+	if !canFailover {
+		span.SetStatus(codes.Ok, "unhealthy volume(s)")
+		log.Info("pod has been labeled as fenced but not all volumes are healthy after node failure, leaving pod running")
```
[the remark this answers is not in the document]
croomes: As above, already added.

controllers/fencer/reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	span.SetAttributes(label.String("name", req.NamespacedName.Name))
+	defer span.End()
+	log := r.log.WithValues("name", req.NamespacedName)
+
+	// Query the object details from the cache and operate on it.
+	val, found := r.cache.Get(req.NamespacedName.String())
+	if !found {
+		log.Error(ErrNodeNotCached, "skipping node, will retry")
+		span.RecordError(ErrNodeNotCached)
+		return ctrl.Result{}, nil
```
darkowlzz, line 184: The log says that it'll retry, but the result has no requeue set. I think that's fine, the log should not say it'll retry because usually we don't know what kind of event we received, in case of delete event, Get fails and we don't do anything. But in case of this specific controller, it only receives generic events, not create, update or delete events. So an event means something changed and we may not ignore that and actually retry.
This makes me think about the consequences of 1 minute cache expiry time. If due to some reason, we retry here or also below where `fenceNode()` is called, and due to some reason we fail to reconcile within a minute, the cache will be empty and we'll not be able to do anything because we'll not have the event data. Having a longer cache expiry and cleanup time doesn't hurt because cache gets updated for any observed change in objects.

internal/pkg/cache/cache.go
```diff
@@ -0,0 +1,47 @@
...
+		Cache: cache.New(expiryInterval, cleanupInterval),
+		log:   log.WithName("ObjectCache"),
+	}
+}
+
+// CacheMiss implements external controller Cache interface.
+// It checks for cache miss with the given object. On cache miss, it updates
+// the cache.
+func (c *Object) CacheMiss(obj client.Object) bool {
+	key := Key(obj.GetName(), obj.GetNamespace())
```
darkowlzz, line 36: Simpler to just use `client.ObjectKeyFromObject()`. 🙂

internal/pkg/cache/cache.go
```diff
@@ -0,0 +1,47 @@
...
+
+	cached, found := c.Get(key)
+	if found && reflect.DeepEqual(cached, obj) {
```
darkowlzz, line 39: Maybe we could use something like [this](https://github.com/kubernetes/apimachinery/blob/10b38829b6211c0d672e14a126479221312d3368/pkg/api/apitesting/roundtrip/roundtrip.go#L308) from apimachinery for semantic deep equal of objects?

internal/pkg/cache/cache.go
```diff
@@ -0,0 +1,47 @@
...
+		c.log.V(5).Info("cache hit", "name", obj.GetName(), "namespace", obj.GetNamespace())
+		return false
+	}
+
+	c.Set(key, obj, 0)
+	c.log.V(5).Info("cache miss", "name", obj.GetName(), "namespace", obj.GetNamespace())
```
darkowlzz, line 45: Maybe it'll be better to be more specific and use "key" to refer to the object attributes here, and use the key directly. Key is a NamespacedName and that has a nice stringer method with "/" in between the values.

croomes: agreed - done

internal/pkg/cache/cache.go
croomes: Good idea - done

internal/pkg/cache/cache.go
croomes: yes!

controllers/node-delete/controller.go
```diff
@@ -67,12 +67,17 @@ func (c Controller) List(ctx context.Context) ([]types.NamespacedName, error) {
...
 		return nil, err
 	}
-	span.SetAttributes(label.Int("count", len(objects)))
+	span.SetAttributes(label.Int("count", len(nodes)))
 	span.SetStatus(codes.Ok, "listed nodes")
-	return storageos.ObjectKeys(objects), nil
+
+	keys := []types.NamespacedName{}
+	for _, n := range nodes {
+		keys = append(keys, types.NamespacedName{Name: n.GetName()})
```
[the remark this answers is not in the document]
croomes: agreed - I've changed to use `client.ObjectKey` in the fencing controller.  I'll do a separate PR for the other controllers.

controllers/fencer/reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	if node.Status.Health != storageosv1.NodeHealthOffline {
+		log.V(5).Info("ignore healthy node")
+		span.SetStatus(codes.Ok, "node healthy")
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("fencing pods on node")
+	if err := r.fenceNode(ctx, node); err != nil {
+		span.RecordError(err)
+		return ctrl.Result{}, fmt.Errorf("failed to fence node: %v", err)
```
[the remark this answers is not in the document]
croomes: In testing we found that the Pod was sometimes getting re-scheduled before the VolumeAttachments were deleted, causing errors to be logged (which would later succeed).  In the last commit I moved the Pod delete to after the VA delete, and returned an error for any VA delete failure.  This will cause the fence operation to requeue until all VAs and the Pod are successfully deleted. 

Previously it was ignoring the VA delete errors as the Pod was no longer available to list the PVCs->VAs if requeued.

This is much better behaviour as there's no point in deleting the Pod if there's still a VA present that will block it restarting.

controllers/fencer/reconciler.go
croomes: I'm now returning an error here, and I've increased the cache expiry to 10 minutes.  An hour might be better?

README.md
```diff
@@ -24,6 +24,13 @@ Endpoint will be updated, but the Service ClusterIP will remain the same.
 See [Shared Volume Contoller](internal/controllers/sharedvolume/README.md) for
 more detail.
 
+### Fencing Controller
+
+The fencing controller performs fast failovers of Pods with StorageOs volumes
```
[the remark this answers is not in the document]
croomes: Done

controllers/fencer/reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Initialize the reconciler with the fencing controller.
+	r.Reconciler.Init(mgr, c,
+		actionv1.WithName("pod-fencer"),
+		actionv1.WithScheme(mgr.GetScheme()),
+		actionv1.WithActionTimeout(25*time.Second),
+		actionv1.WithActionRetryPeriod(5*time.Second),
```
darkowlzz, line 134: Maybe these need to be made configurable?

main.go
```diff
@@ -81,17 +93,20 @@ func main() {
...
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.StringVar(&apiSecretPath, "api-secret-path", "/etc/storageos/secrets/api", "Path where the StorageOS api secret is mounted.  The secret must have \"username\" and \"password\" set.")
 	flag.StringVar(&apiEndpoint, "api-endpoint", "storageos", "The StorageOS api endpoint address.")
-	flag.DurationVar(&apiPollInterval, "api-poll-interval", 5*time.Second, "Frequency of StorageOS api polling.")
+	flag.DurationVar(&volumePollInterval, "volume-poll-interval", 5*time.Second, "Frequency of StorageOS volume polling.")
+	flag.DurationVar(&nodePollInterval, "node-poll-interval", 5*time.Second, "Frequency of StorageOS node polling.")
+	flag.DurationVar(&volumeExpiryInterval, "volume-expiry-interval", time.Minute, "Frequency of cached StorageOS volume re-validation.")
+	flag.DurationVar(&nodeExpiryInterval, "node-expiry-interval", 15*time.Minute, "Frequency of cached StorageOS node re-validation.")
```
darkowlzz, line 107: The cache is set to expire at 15 minutes. Since we don't depend on cache flush for anything, we can increase the default even more if we want to.

croomes: Let's set to 1 hour

controllers/fencer/reconciler.go
croomes: Yes, we should - thanks.  The default values look good.

darkowlzz:
Rebased.

darkowlzz:
Rebased.

controllers/fencer/reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+	// Initialize the cache.
+	r.cache = cache.New(r.expiryInterval, r.cleanupInterval)
+
+	// Create an event handler that uses the cache to make reconciliation
+	// decisions.
+	eventHandler := handler.NewEnqueueRequestFromCache(r.cache)
+
+	// Don't allow an overly-agressive interval to break the backend.
+	if r.pollInterval < minPollInterval {
+		r.log.Info("resetting poll interval to minimum", "interval", minPollInterval)
```
mhmxs, line 100: I guess there is a missing line here, something like `r.pollInterval = minPollInterval`

mhmxs: Maybe if this is something which could break the backend some test should be nice to cover it. What do you think?

darkowlzz: You're right, an assignment is missing.
We don't really know what the control plane is capable of handling. So this is just a vague minimum interval for safety.
I think this conditional can be properly tested by moving this into the constructor `NewReconciler`.

croomes:
Just had a look through and I'm happy with the updates.  Feel free to merge when ready.

mhmxs (approved):

PR #37, opened 2021-03-22
Move interfaces to callers

croomes:
This is an internal-only change to avoid a circular dependency when writing tests that use the mock control plane API client.  Having the caller define the interface it requires follows go best practices.

This is split out from a larger PR that uses the mock control plane API client for its tests.

Also fixes a linter failure.


darkowlzz (approved):
Makes sense. Thanks. LGTM!

PR #38, opened 2021-03-22
API token reset failures should trigger re-authentication directly

croomes:
Fix a possible deadlock when the API token refresh goroutine and a controller try to reset the API connection.  Instead of both sending to the same channel, the API refresh should re-authenticate directly.

Branched off https://github.com/storageos/api-manager/pull/37, needed for the test.


internal/pkg/storageos/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+			timeout:  1050 * time.Millisecond,
+			expect: func(ctx context.Context, m *mocks.MockControlPlane) {
+				httpResp := &http.Response{
+					Header: http.Header{
+						"Authorization": []string{"Bearer newtoken"},
+					},
+					Body: ioutil.NopCloser(bytes.NewReader([]byte(""))),
+				}
+				m.EXPECT().RefreshJwt(gomock.Any()).Return(api.UserSession{}, nil, errRefresh).Times(1)
+				m.EXPECT().AuthenticateUser(gomock.Any(), api.AuthUserData{Username: "username", Password: "password"}).Return(api.UserSession{}, httpResp, nil).Times(1)
```
darkowlzz, line 77: Let's use constants for all the "username" and "password" literals in all the test cases, reducing the chances of typo based failures.

internal/pkg/storageos/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	defer os.RemoveAll(secretPath)
+	if err := ioutil.WriteFile(filepath.Join(secretPath, "username"), []byte("username"), 0644); err != nil {
+		t.Fatalf("failed to write username file: %v", err)
+	}
+	if err := ioutil.WriteFile(filepath.Join(secretPath, "password"), []byte("password"), 0644); err != nil {
+		t.Fatalf("failed to write password file: %v", err)
+	}
```
darkowlzz, line 33: Since this is a time based test, it'd be better to run them in parallel. How about moving the above in the `t.Run()`, to avoid tests from conflicting with each other, and enable `t.Parallel()`?

darkowlzz (commented):
Thanks, LGTM!
Suggested minor improvements in the tests.

PR #40, opened 2021-03-30
Set Pod scheduler

croomes:
Adds a mutating webhook to set the StorageOS scheduler on pods with StorageOS PVCs.


main.go
```diff
@@ -86,6 +107,15 @@ func main() {
...
+	flag.StringVar(&schedulerName, "scheduler-name", "storageos", "Name of the Pod scheduler to use for Pods with StorageOS volumes.  Set to an empty value to disable setting the Pod scheduler.")
+	flag.StringVar(&webhookServiceName, "webhook-service-name", "storageos-scheduler-webhook", "Name of the webhook service.")
+	flag.StringVar(&webhookServiceNamespace, "webhook-service-namespace", "", "Namespace of the webhook service.  Will be auto-detected or value of -namespace unset.")
+	flag.StringVar(&webhookSecretName, "webhook-secret-name", "storageos-webhook-secret", "Name of the webhook secret.")
+	flag.StringVar(&webhookSecretNamespace, "webhook-secret-namespace", "", "Namespace of the webhook secret.  Will be auto-detected or value of -namespace unset.")
+	flag.StringVar(&webhookConfigMutatingName, "webhook-config-mutating", "storageos-mutating-webhook", "Name of the mutating webhook configuration.")
+	flag.StringVar(&webhookMutatePodsPath, "webhook-mutate-pods-path", "/mutate-pods", "URL path of the Pod mutating webhook.")
+	flag.DurationVar(&webhookCertRefreshInterval, "webhook-cert-refresh-interval", 3*30*24*time.Hour, "Frequency of webhook certificate refresh.")
```
darkowlzz, line 118: We decided to do this often, maybe about every 30 minutes or less, to reduce the amount of time multiple manager could be out of sync, if that ever happens.

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	go.opentelemetry.io/otel v0.15.0
 	go.uber.org/zap v1.15.0
 	k8s.io/api v0.20.2
+	k8s.io/apiextensions-apiserver v0.20.2
```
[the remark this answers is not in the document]
croomes: Agreed, was only used in a test and could be removed.

main.go
croomes: Set to 30 minutes

go.mod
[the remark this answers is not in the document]
croomes: done

controllers/suite_test.go
```diff
@@ -8,7 +8,9 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"k8s.io/client-go/kubernetes/scheme"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
```
darkowlzz, line 11: I don't think we need this admission v1beta1 API as well.

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,25 @@
+
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
```
darkowlzz, line 3: We can use v1 API here as well.

main.go
```diff
@@ -86,6 +107,15 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&namespace, "namespace", "", "Namespace that the StorageOS components, including api-manager, are installed into.  Will be auto-detected if unset.")
+	flag.StringVar(&schedulerName, "scheduler-name", "storageos-scheduler", "Name of the Pod scheduler to use for Pods with StorageOS volumes.  Set to an empty value to disable setting the Pod scheduler.")
+	flag.StringVar(&webhookServiceName, "webhook-service-name", "storageos-mutating-webhook", "Name of the webhook service.")
```
darkowlzz, line 112: Maybe we should have a more generic value for this. The same service will be shared with validating webhook. Similar to the secret name, `storageos-webhook-service` or just `storageos-webhook` should be better.

darkowlzz (commented):
Overall, looks very good. Left a few inline comments.

controllers/admission/scheduler/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Webhook server
+
+The Pod Scheduler admission controller runs as a webhook within api-manager.
+The webhook server is shared by other admission controllers and uses a
+self-signed certificate.
+
+Certificates are rotated automatically and stored in a secret.  Multiple
+instances of the api-manager shared the same certificate and will check for
+updates periodically, configured with the `-webhook-cert-refresh-interval` flag.
```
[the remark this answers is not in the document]
croomes: ok, we can update when we add cert-manager support.

main.go
croomes: Agreed, I've changed the secret and service name to be both `storageos-webhook`

config/webhook/service.yaml
```diff
@@ -2,8 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: webhook-service
-  namespace: system
+  name: storageos-webhook-service
```
darkowlzz, line 5: It's just `storageos-webhook` now :slightly_smiling_face:

croomes: Grr...  missed that - thanks

darkowlzz (approved):
LGTM!
There's one document to be updated https://github.com/storageos/api-manager/pull/40/files#r607888042
We can squash the commits and merge it.
Thanks for all the work.

PR #41, opened 2021-04-09
Allow multiple mutators for a single admission controller

croomes:
This allows an admission controller to run multiple mutators for different tasks.

For Pod creation, this could be to add the StorageOS scheduler and to set a failure mode or failure mode label that should be applied to all of the Pod's PVCs.

This pattern will be used for PVC create where we will have separate mutators for encryption and labels.


controllers/pod-mutator/scheduler/scheduler.go
```diff
@@ -101,9 +65,9 @@ func (p *PodSchedulerSetter) mutatePodsFn(ctx context.Context, pod *corev1.Pod,
 
 	// Skip mutation if the pod annotation has false schedule annotation.
 	if val, exists := pod.ObjectMeta.Annotations[p.SchedulerAnnotationKey]; exists {
-		boolVal, err := strconv.ParseBool(val)
+		enabled, err := strconv.ParseBool(val)
 		// No error in parsing and the value is false, skip the pod.
-		if err == nil && !boolVal {
+		if err == nil && !enabled {
 			log.V(4).Info("pod has storageos.com/scheduler=false annotaion, skipping")
```
avestuk, line 71:
```suggestion
			log.V(4).Infof("pod has %s=false annotaion, skipping", p.SchedulerAnnotationKey)
```

Does it make sense to use the const you have defined already? I suppose there's not much chance of it changing. Really I'm just curious about whether you chose to not use the const.

controllers/pod-mutator/controller.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	// Get the request namespace. This is needed because sometimes the decoded
+	// pod object lacks namespace info.
+	namespace := req.AdmissionRequest.Namespace
+
+	// Create a copy of the pod to mutate.
+	copy := pod.DeepCopy()
```
avestuk, line 56: Why do you have to create a copy of the pod as opposed to just modifying the `pod` object directly and sending that back?

avestuk (approved):

controllers/pod-mutator/README.md
```diff
@@ -1,41 +1,35 @@
...
+instances of the api-manager share the same certificate and will check for
 updates periodically, configured with the `-webhook-cert-refresh-interval` flag.
 
 Certificates are valid for 1 year and re-issued after 6 months.  The
 `-webhook-cert-refresh-interval` should be kept to run frequently (default
 `30m`) as restarting the api-manager will reset the refresh timer.
 
+## Disabling
+
+It is not possible to disable the Webhook server.  Instead, disable thegla
```
mhmxs, line 30: `thegla` looks a typo here.

controllers/pod-mutator/scheduler/README.md
```diff
@@ -0,0 +1,33 @@
...
+not have the `storageos.com/scheduler=false` annotation set on the Pod.
+
+## Trigger
+
+All Pod create events are evaluated.
+
+Pods that have the annotation `storageos.com/scheduler=false` set are ignored.
+
+The Pod's volumes are enumerated, and if at least one volume is backed by
+StorageOS, the SchedulerName will be changed to use the StorageOS Pod scheduler.
```
mhmxs, line 14: SchedulerName should be `highlighted`

controllers/pod-mutator/controller.go
darkowlzz: This code is based on https://github.com/storageos/cluster-operator/blob/v2.3.4/internal/pkg/admission/scheduler/handler.go#L64 where we used to pass the pod object to `PatchResponse` at the very end of the function to return a response. In the latest webhook package of controller runtime, `PatchResponse` is replaced with `PatchResponseFromRaw` which accepts the raw object in the request. So, we avoid doing a deep copy here.

controllers/pod-mutator/scheduler/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Tunables
+
+Default values work well when the api-manager is installed by the
+cluster-operator.  They should only be changed under advisement from StorageOS
+support.
+
+`-scheduler-name` is the name of the scheduler extender to set in the Pod's
+`SchedulerName` field.  If set to an empty string, no updates will be made. When
+set, the scheduler name must match the name of the scheduler extender configured
+in the StorageOS cluster-operator. (default "storageos-scheduler)".
```
darkowlzz, line 33:
```suggestion
in the StorageOS cluster-operator. (default "storageos-scheduler").
```

controllers/pod-mutator/scheduler/scheduler.go
```diff
@@ -37,59 +34,26 @@ type PodSchedulerSetter struct {
...
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
-}
-
-// mutatePodFn mutates a given pod with a configured scheduler name if the pod
-// is associated with volumes managed by the configured provisioners.
-func (p *PodSchedulerSetter) mutatePodsFn(ctx context.Context, pod *corev1.Pod, namespace string) error {
+// MutatePod mutates a given pod with a configured scheduler name if the pod is
+// associated with volumes managed by the configured provisioners.
+//
+// Errors returned here will block creation of the Pod.
```
darkowlzz, line 55: May not be accurate. It depends on the webhook configuration if the failure is ignored or respected.

darkowlzz:
The implementation looks good to me. Left some inline comments about minor improvements.

controllers/pod-mutator/scheduler/scheduler.go
croomes: yeah, it probably does.  With `logr` they recommend against vars in the message and encourage k=v pairs instead, but that doesn't work well here.  Changed.

controllers/pod-mutator/controller.go
croomes: Removed the `DeepCopy()`

controllers/pod-mutator/scheduler/scheduler.go
croomes: :+1: Changed wording

controllers/test_helpers.go
```diff
@@ -0,0 +1,45 @@
+package controllers
```
mhmxs, line 1: In this way test related codebase looks like production code. I suggest to refactor the code to ensure test functions are only available in tests. If it is not possible a common solution to create a `controllertest` package for separation.

controllers/test_helpers.go
```diff
@@ -0,0 +1,45 @@
...
+	if m.error {
+		return errors.New("error")
+	}
+	if obj.Annotations == nil {
+		obj.Annotations = make(map[string]string)
+	}
+	obj.Annotations[m.key] = m.value
+	return nil
+}
+func (m testMutator) MutatePVC(ctx context.Context, obj *corev1.PersistentVolumeClaim, namespace string) error {
```
mhmxs, line 36: Missing new line :)

mhmxs (approved):

controllers/test_helpers.go
mhmxs: Sorry, my bad, this package contains only tests :D

darkowlzz (approved):
Other than the remaining minor comments and squashing the commits, this LGTM!
Thanks.

controllers/test_helpers.go
croomes: done, thanks

PR #42, opened 2021-04-09
Add PVC mutating controller

croomes:
Adds the controller to handle mutating PVC create requests.  There will be a separate operator PR to create the MutatingWebhookConfiguration, but this can come later.

No mutators have been added at this point.

I'll re-target to master after https://github.com/storageos/api-manager/pull/41 merges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,15 @@ StorageOS cluster when the namespace has been removed from Kubernetes.
 See [Namespace Delete Controller](controllers/namespace-delete/README.md) for
 more detail.
 
+## Admission Controllers
+
+Admission controllers intercept requests to the Kubernetes API prior to the
+object being persisted, but after authentication and authorisation.
+
+Mutating admission controllers run first and can modify the object, followed by
+Validating admission controllers.  Both can cause the Kubernetes API to reject
+the request.
+
 ### Pod Mutator Admission Controller
 
 The Pod Mutator is a mutating admission controller that modifies Pods during the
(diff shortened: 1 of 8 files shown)


avestuk (approved):

mhmxs (approved):

README.md
```diff
@@ -44,7 +44,15 @@ more detail.
 The Pod Mutator is a mutating admission controller that modifies Pods during the
 create process.
 
-See [Namespace Delete Controller](controllers/pod-mutator/README.md) for
+See [Pod Mutator Admission Controller](controllers/pod-mutator/README.md) for
```
darkowlzz, line 47: Left a comment in the other PR to fix this where it's introduced.

config/webhook/manifests.yaml
```diff
@@ -23,3 +23,20 @@ webhooks:
     - CREATE
     resources:
     - pods
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-pvcs
+  failurePolicy: Ignore
```
darkowlzz, line 32: I think we still need to decide if this should be ignore or fail. Mostly likely fail because we don't want to create a misconfigured volume.

controllers/pvc-mutator/README.md
```diff
@@ -0,0 +1,62 @@
...
+- TODO(croomes): List of mutators should be added here.
+
+## Webhook server
+
+The admission controller runs as a webhook within api-manager.
+The webhook server uses a self-signed certificate.
+
+Certificates are rotated automatically and stored in a secret.  Multiple
+instances of the api-manager share the same certificate and will check for
+updates periodically, configured with the `-webhook-cert-refresh-interval` flag.
```
darkowlzz, line 20: This seems repetitive. We have similar paragraphs in pod mutator too. It'll increase docs maintenance work.

darkowlzz: The following paragraphs seem to be the same as in the pod mutator. Maybe we should have a common docs for all the mutators.

controllers/pvc_mutator_test.go
```diff
@@ -0,0 +1,207 @@
...
+					},
+				},
+				VolumeMode: &volumeMode,
+			},
+		}
+	}
+
+	ctx := context.Background()
+
+	Context("When the PVC Mutator has no mutators", func() {
```
darkowlzz, line 107: Looks similar to the tests that we have for pod mutators. The only difference is in the mutator interface's mutate function signature. I guess we can avoid all these with a common package that provides such mutation/validation chaining by accepting something generic like a client.Object and the core logic is tested just once.

config/webhook/manifests.yaml
croomes: Yes, good point.  We don't want to silently create a volume without encryption if it was requested.  Changed.

controllers/pvc-mutator/README.md
croomes: Agreed.  I've moved the webhook docs into the main `README`.  It should be a lot less repetitive now.

controllers/pvc_mutator_test.go
croomes: Yes, that would be better.  Once we add a common package we can refactor, but for now I think testing this explicitly is good.

config/webhook/manifests.yaml
croomes: Setting the default to fail has the side-effect that the e2e tests for the other controllers fail unless the webhook is started.  I'd rather not have the e2e tests depend on other controllers so setting the default back to ignore.

Longer-term we should create a new envtest for each controller with just the config that it requires.  This means moving envtest setup from `BeforeSuite` to `BeforeEach`.

darkowlzz (approved):
Commits need to squashing.
Everything else LGTM!
Thanks.

PR #43, opened 2021-04-13
Migrate to kubebuilder v3

darkowlzz:
- Updates the makefile to be based on the kubebuilder v3 makefile.
- Update kubebuilder, controller-tools and kustomize dependencies. All
the binaries are downloaded into ./bin dir. And all the test related
binaries are downloaded in testbin. Old assets in
/usr/local/kubebuilder/ can be deleted. All the dependencies are
maintained within the project directory.
- envtest related tools are downloaded when the tests are run. All build
and test dependencies are self-contained now.
- Update all the generated code and configs.
- make help shows a list of all the targets with description.

NOTE: We have already been using most of the kubebuilder v3 dependencies
with the latest stable controller-runtime. This change moves the project scaffold,
code generation and testing tools to the versions used in kubebuilder v3.
Although kubebuilder v3 doesn't officially supports go 1.16 yet, since we don't
have any custom resources API generation in this project, we are good to use
go 1.16.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,15 @@
 # Image URL to use all building/pushing image targets
 IMG ?= storageos/api-manager:test
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 
-OS=$(shell go env GOOS)
-ARCH=$(shell go env GOARCH)
-KUBEBUILDER_VERSION=2.3.1
-DEFAULT_KUBEBUILDER_PATH=/usr/local/kubebuilder/bin
+KUBEBUILDER_VERSION=3.0.0-rc.0
+CONTROLLER_RUNTIME_VERSION=0.8.3
+CONTROLLER_TOOLS_VERSION=0.4.1
+KUSTOMIZE_VERSION=3.8.7
+# Set the shell as a workaround for the issue
+# https://github.com/operator-framework/operator-sdk/issues/4203
+SHELL=bash
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -16,86 +17,114 @@ else
...
+	cd - && mv bin/cmd $(KUBEBUILDER) ;\
 	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
-endif
+
+CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
+controller-gen: ## Download controller-gen locally if necessary.
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)
```
mhmxs, line 107: Could you please create a variable for this version number?

Makefile
```diff
@@ -16,86 +17,114 @@ else
...
+
+KUSTOMIZE = $(shell pwd)/bin/kustomize
+kustomize: ## Download kustomize locally if necessary.
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
```
mhmxs, line 111: Could you please create a variable for this version number?

mhmxs (changes requested):

Makefile
darkowlzz: Yes, that can be done. I didn't change it because this is what kb v3 generates and was thinking to make as little change as possible.

mhmxs: agree, thanks

PROJECT
```diff
@@ -1,4 +1,7 @@
 domain: storageos.com
+layout:
+- go.kubebuilder.io/v3
+projectName: api-manager
```
croomes, line 4: Doesn't adding `projectName` cause problems when generating new controllers?  Something to do with the resource naming...  I'll try.

darkowlzz: I don't think so. And I think our controller package directory structure is incompatible with what the controller-gen supports.

croomes: It works well as you have it:

```
./bin/kubebuilder create api --group "" --version v1 --kind Volume
```
Creates:
```
GroupVersion = schema.GroupVersion{Group: "storageos.com", Version: "v1"}
```
:+1:

croomes (approved):
Thanks!

mhmxs (approved):

PR #44, opened 2021-04-15
Rename leader election config map

mhmxs:
Replace hard coded randomish name to a proper one.

### DIFF
--- a/main.go
+++ b/main.go
@@ -190,7 +190,7 @@ func main() {
 		MetricsBindAddress: metricsAddr,
 		Port:               9443,
 		LeaderElection:     enableLeaderElection,
-		LeaderElectionID:   "api-manager-lease",
+		LeaderElectionID:   "storageos-api-manager-lease",
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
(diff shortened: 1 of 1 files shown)


croomes (approved):
Thanks!

main.go
```diff
@@ -190,7 +190,7 @@ func main() {
 		MetricsBindAddress: metricsAddr,
 		Port:               9443,
 		LeaderElection:     enableLeaderElection,
-		LeaderElectionID:   "d73494fd.storageos.com",
+		LeaderElectionID:   "api-manager-lease",
```
darkowlzz, line 193: With this name change, we are losing the information that this belongs to storageos. Since our current default deployment is in kube-system namespace, along with many other system resources, it'll make it harder to identify that this belongs to storageos, if we want to find that out for some reason.
Maybe `api-manager-lease.storageos.com` would be better to keep the ownership info in the name?
Just an opinion.

croomes: Good point, but my preference would be to use the `storageos-` prefix we use for other resources, so something like `storageos-api-manager-lease`.

darkowlzz (approved):
LGTM!
Please squash the commits and merge.
Thanks.

PR #45, opened 2021-04-15
Add support for volume encryption

croomes:
Adds a mutator to intercept PVC create requests and generates encryption keys and stores them in secrets if encryption has been requested.  The reference to the secret is stored in annotations on the PVC.


controllers/pvc-mutator/encryption/keygen.go
```diff
@@ -0,0 +1,185 @@
...
+// which may be deleted and then the PV reused.  Since the volume hasn't been
+// provisioned yet we don't have a reference for it, so generate a unique
+// identifier to use instead.
+func VolumeSecretName() string {
+	return fmt.Sprintf("%s-%s", VolumeSecretNamePrefix, uuid.New().String())
+}
+
+// isEnabled returns true if the key is set in the kv map and its value is true.
+func isEnabled(key string, kv map[string]string) bool {
+	if kv == nil {
```
mhmxs, line 172:
```
Map types are reference types, like pointers or slices, and so the value of m above is nil; it doesn't point to an initialized map. A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic
```
So `kv` is a non nil pointer, so nil check doesn't necessary (or makes any sense)

controllers/pvc-mutator/encryption/keys/manager.go
```diff
@@ -0,0 +1,174 @@
...
+// Ensure that a secret exists at volKeyRef, creating it with valid keys if
+// needed.  If a secret already exists, it does not verify validity.
+func (m *KeyManager) Ensure(ctx context.Context, nsKeyRef client.ObjectKey, volKeyRef client.ObjectKey) error {
+	// Return immediately if the volume key already exists, or an error if
+	// another error (e.g. permission denied).
+	err := m.client.Get(ctx, volKeyRef, &corev1.Secret{})
+	if err == nil {
+		return nil
+	}
+	if !apierrors.IsNotFound(err) {
```
mhmxs, line 42: This if statement technically belongs to the previous. In that case i ususally prefer `else if`.

controllers/pvc-mutator/encryption/keygen.go
croomes: Removed and added test

controllers/pvc-mutator/encryption/keys/manager.go
croomes: I'm not sure I agree with this one.  Where an error handling case results in a return, then there's no need to add an `else`.  I find this easier to read, especially when not adding a space between lines to indicate that it's a related block.

controllers/pvc-mutator/encryption/keygen.go
```diff
@@ -0,0 +1,185 @@
...
+	// should only be added to StorageOS PVCs.
+	if !isEnabled(s.enabledLabel, pvc.GetLabels()) {
+		log.V(4).Info("pvc does not have storageos.com/encryption=true label, skipping")
+		return nil
+	}
+
+	// Do not allow secrets from another namespace to be referenced.  We assume
+	// that if the user can create PVCs in the namespace then they should be
+	// able to read volume secrets from that namespace.  We can't assume the
+	// same for other namespaces.
```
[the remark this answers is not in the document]
croomes: Yes, that may be needed in the future.  At the moment I'd rather restrict the feature rather than relaxing security constraints.  I think we need to give it a bit more thought and document it properly before allowing it.

controllers/pvc-mutator/encryption/keys/manager.go
mhmxs: I can agree. At the concrete situation both ifs are working on the same data. The shortest/lower_scope version looks this:
```
if err := m.client.Get(ctx, volKeyRef, &corev1.Secret{}); err == nil {
    return nil
} else if !apierrors.IsNotFound(err) {
    return err
}
```
I can Go further :D 
```
if err := m.client.Get(ctx, volKeyRef, &corev1.Secret{}); err == nil || !apierrors.IsNotFound(err) {
    return err
}
```

I mark this resolved, because your view also makes sense.

mhmxs (approved):

controllers/pvc-mutator/encryption/keys/manager.go
croomes: oh, I like your last suggestion!  Changed.

mhmxs: I found an other place: https://github.com/storageos/api-manager/pull/45/files#diff-de1ea10f4b2cc388d0a7f3bdbc4fd1dfedacb0d1353361dcaf7e193426acff18R54

mhmxs (approved):

controllers/pvc-mutator/encryption/keygen_test.go
```diff
@@ -0,0 +1,388 @@
...
+				t.Errorf("expected %s annotation to be set to %s, got %s", SecretNameAnnotationKey, tc.wantSecretNameAnnotation, nameRef)
+			}
+			if !tc.wantSecretNameAnnotationGenerated && tc.wantSecretNameAnnotation == "" && nameRef != "" {
+				t.Errorf("expected %s annotation to be unset, got %s", SecretNameAnnotationKey, nameRef)
+			}
+
+			// Check namespace ref annotation.
+			nsRef := annotations[SecretNamespaceAnnotationKey]
+			switch tc.wantSecretNamespaceAnnotation == "" {
+			case true:
```
darkowlzz, line 177: Maybe
```go
if tc.wantSecretNamespaceAnnotation == "" {
  ...
} else {
  ...
}
```

controllers/pvc-mutator/encryption/keygen_test.go
```diff
@@ -0,0 +1,388 @@
...
+		Provisioner: "foo-provisioner",
+	}
+
+	testNamespace := "default"
+
+	testcases := []struct {
+		name                              string
+		namespace                         string
+		notStos                           bool
+		betaAnnotation                    bool
```
darkowlzz, line 56: Looks like none of the test cases use `notStos` and `betaAnnotation`. It'd be nice to test them as well, to ensure that the mutation doesn't fail for them due to some reason.

controllers/pvc-mutator/encryption/keys/manager.go
```diff
@@ -0,0 +1,173 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/storageos/api-manager/internal/pkg/crypto"
+)
+
+var (
+	// ErrNoKeyInSecret is returned is a secret was expected to contain an
```
darkowlzz, line 18:
```suggestion
	// ErrNoKeyInSecret is returned when a secret was expected to contain an
```

controllers/pvc-mutator/encryption/keys/manager.go
```diff
@@ -0,0 +1,173 @@
...
+func New(client client.Client) *KeyManager {
+	return &KeyManager{client: client}
+}
+
+// Ensure that a secret exists at volKeyRef, creating it with valid keys if
+// needed.  If a secret already exists, it does not verify validity.
+func (m *KeyManager) Ensure(ctx context.Context, nsKeyRef client.ObjectKey, volKeyRef client.ObjectKey, nsSecretLabels map[string]string, volSecretLabels map[string]string) error {
+	// Return immediately if the volume key already exists, or an error if
+	// another error (e.g. permission denied).
+	if err := m.client.Get(ctx, volKeyRef, &corev1.Secret{}); err == nil || !apierrors.IsNotFound(err) {
```
darkowlzz, line 39: Returning err if err is nil. Doesn't look right. :thinking:

darkowlzz: Just saw https://github.com/storageos/api-manager/pull/45#discussion_r614680821 . idk if it makes it hard to read.

controllers/pvc-mutator/encryption/keys/manager.go
```diff
@@ -0,0 +1,173 @@
...
+	secret := m.secret(nsKeyRef, data, labels)
+	if err := m.client.Create(ctx, secret); err != nil {
+		return "", err
+	}
+
+	// Wait for secret to be created.  There's not much value in making these
+	// configurable.  The create should be almost instant.
+	ticker := time.NewTicker(50 * time.Millisecond)
+	timeout := time.After(2 * time.Second)
+	defer ticker.Stop()
```
darkowlzz, line 81: It'll be better to use an uncached client for such things.

controllers/pvc-mutator/encryption/keygen_test.go
mhmxs: `notStos` is useful and used one of the PR-s opened on top of this!

controllers/pvc-mutator/encryption/keys/manager.go
croomes: Agreed, putting it back.

PR #46, opened 2021-04-17
Add flag and support for cert validity

mhmxs:
Operator toolkit support cert validity so this PR introduces the option in api manager.


croomes (approved):
Thanks!

PR #48, opened 2021-04-19
Copy StorageClass params to labels during PVC creation

mhmxs:
To avoid inconsistency in labels during regular label sync this new PVC mutator ensures all the params defined in StorageClass has it's equivalent pair as label.


controllers/pvc-mutator/storage-class-label-sync/sync.go
```diff
@@ -0,0 +1,55 @@
+package storageclasslabelsync
```
croomes, line 1: This is a bit of a mouthful - how about `sclabel`?  This matches better with existing packages: `pvclabel`, `nodelabel`, `nodedelete` and `nsdelete`.

Same with the directory name, I think `sc-label` would be better.

controllers/pvc-mutator/storage-class-label-sync/sync.go
```diff
@@ -0,0 +1,55 @@
...
+	return &StorageClassParamsLabelSetter{
+		Client: k8s,
+		log:    ctrl.Log,
+	}
+}
+
+// MutatePVC mutates a given pvc with labels based on,
+// StoraeClass params.
+//
+// Errors returned here will block creation of the PVC.
```
croomes, line 32: I think this would be better - I'm going to change this in the encryption PR too:
```
// Errors returned here may block creation of the PVC, depending on the
// FailurePolicy set in the webhook configuration.
```

croomes (commented):
Looks good, thanks!

I'd just suggest the rename and adding an e2e test.

controllers/pvc-mutator/storage-class-label-sync/sync.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	// Set labels on the PVC based on params
+	if pvc.Labels == nil {
+		pvc.Labels = make(map[string]string)
+	}
+	for k, v := range storageClass.Parameters {
+		v := v
+		pvc.Labels[k] = v
+	}
```
croomes, line 51: If a user has specified their own PVC labels, then don't overwrite them.  SC params are defaults that a user can override with PVC labels.

controllers/pvc-mutator/storage-class-label-sync/sync.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// MutatePVC mutates a given pvc with labels based on,
+// StoraeClass params.
+//
+// Errors returned here will block creation of the PVC.
+func (s *StorageClassParamsLabelSetter) MutatePVC(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace string) error {
+	log := s.log.WithValues("pvc", client.ObjectKeyFromObject(pvc).String())
+	log.V(4).Info("received pvc for mutation")
```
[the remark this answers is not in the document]
mhmxs: I prefer `.WithName()` for both.

controllers/pvc-mutator/storage-class-label-sync/sync.go
mhmxs: Does it true for `EncryptionKeySetter.MutatePVC()` too? At the moment i don't see the reason if not :)

controllers/pvc-mutator/sc-label/sync.go
```diff
@@ -0,0 +1,70 @@
...
+	client.Client
+	log logr.Logger
+}
+
+// NewLabelSetter returns a new PVC label sync mutating admission
+// controller.
+func NewLabelSetter(k8s client.Client) *LabelSetter {
+	return &LabelSetter{
+		Client: k8s,
+		log:    ctrl.Log,
```
croomes: I'd set the logger name here, rather than on on L36:

```
		log:    ctrl.Log.WithName("sc-param-sync"),
```
This format matches the other controllers, e.g. `node-label-sync`, `volume-label-sync`.

controllers/pvc-mutator/storage-class-label-sync/sync.go
croomes: Yes, I've made the same wording change for encryption.

controllers/pvc-mutator/sc-label/sync.go
```diff
@@ -0,0 +1,69 @@
...
+	storageClass, err := provisioner.StorageClassForPVC(s.Client, pvc)
+	if err != nil {
+		return err
+	}
+
+	// Set labels on the PVC based on parameters
+	if pvc.Labels == nil {
+		pvc.Labels = make(map[string]string)
+	}
+	for k, v := range storageClass.Parameters {
```
croomes, line 59: As discussed, let's only copy params prefixed by `storageos.com/`.  We should also validate that the value is in proper label format and ignore (don't error) if not.

controllers/pvc-mutator/sc-label/sync.go
```diff
@@ -0,0 +1,69 @@
...
+	log := s.log.WithValues("pvc", client.ObjectKeyFromObject(pvc).String())
+	log.V(4).Info("received pvc for mutation")
+
+	// Skip mutation if the PVC is not provisioned by StorageOS
+	provisioned, err := provisioner.IsProvisionedPVC(s.Client, *pvc, namespace, provisioner.DriverName)
+	if err != nil {
+		return errors.Wrap(err, "failed to check pvc provisioner")
+	}
+	if !provisioned {
+		log.V(4).Info("pvc does not provisioned by StorageOS, skipping")
```
avestuk, line 45:
```suggestion
		log.V(4).Info("pvc was not provisioned by StorageOS, skipping")
```

controllers/pvc-mutator/sc-label/sync_test.go
```diff
@@ -0,0 +1,135 @@
...
+			namespace:  testNamespace,
+			labels:     map[string]string{"param": "overwrited"},
+			wantLabels: map[string]string{"param": "overwrited"},
+		},
+		{
+			name:       "has extra label",
+			namespace:  testNamespace,
+			labels:     map[string]string{"extra": "value"},
+			wantLabels: map[string]string{"param": "value", "extra": "value"},
+		},
```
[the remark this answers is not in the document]
mhmxs: I can't totally agree. The StorageClass is loaded by one of the dependencies of the function here: https://github.com/storageos/api-manager/blob/e54f0f19f42cf66db09484e2ad7dad38a7fcb3ea/controllers/pvc-mutator/sc-label/sync.go#L50

In my view testing a side effect is not always the best thing to do. This function doesn't do anything with storage class selection. I more suggest to cover `StorageClassForPVC` with tests instead. What do you advice?

controllers/pvc_sclabel_test.go
```diff
@@ -0,0 +1,176 @@
...
+				},
+				VolumeMode: &volumeMode,
+			},
+		}
+	}
+
+	ctx := context.Background()
+
+	Context("When the PVC Mutator has no mutators", func() {
+		SetupPVCSCLabelTest(ctx)
```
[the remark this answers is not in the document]
mhmxs: `is always configuring the mutator, so I think the test name is incorrect?` my bad, I copied code from keygen test without a brain :D I will fix both in this PR.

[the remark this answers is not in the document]
mhmxs: `I'm not sure I understand the test cases` because the creation of time i thought this is an integration test, so unit tests cover cases, this tests integrity. I'm working on more test cases.

controllers/pvc-mutator/sc-label/sync_test.go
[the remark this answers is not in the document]
croomes: I agree, but can you make sure that `provisioner.StorageClassForPVC()` tests for returning the correct SC when a default is used?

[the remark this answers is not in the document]
mhmxs: I was looking the wrong file. Seems it has been already covered: https://github.com/storageos/api-manager/blob/master/internal/pkg/provisioner/storageclass_test.go#L16

mhmxs:
This change is not need any more.

PR #50, opened 2021-04-20
Add support for failure-mode label

croomes:
Adds the ability to set a volume's failure mode behaviour by updating the PVC's `storageos.com/failure-mode` label.

Also updates the go-api dependency to `v2.4.0`.


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.7.1
-	github.com/storageos/go-api/v2 v2.3.1-0.20210211170707-7520f2fe0974
+	github.com/storageos/go-api/v2 v2.3.1-0.20210420150320-a30cf41359d2
```
croomes, line 16: I had trouble setting to `v2.4.0` so used the commit ref for the tag instead.  I might have something cached as I'd imported from a branch named `v2.4.0` before I tagged.

internal/pkg/storageos/volume_labels.go
```diff
@@ -162,3 +180,77 @@ func (c *Client) EnsureReplicas(ctx context.Context, key client.ObjectKey, desir
...
+
+// EnsureFailureMode ensures that the desired failure mode has been applied to
+// the StorageOS volume.
+func (c *Client) EnsureFailureMode(ctx context.Context, key client.ObjectKey, desired string) error {
+	funcName := "ensure_failure_mode"
+	start := time.Now()
+	defer func() {
+		metrics.Latency.Observe(funcName, time.Since(start))
+	}()
+	observeErr := func(e error) error {
```
mhmxs, line 192: Just an idea. What if we give some name to error in the signature and create a `defer` which checks returned error. If not nil then `increment`. With this helper the rest of the code should able to `return err` like any other regular code, while we have the incrementation too.

internal/pkg/storageos/volume_labels.go
```diff
@@ -162,3 +180,77 @@ func (c *Client) EnsureReplicas(ctx context.Context, key client.ObjectKey, desir
...
+
+	vol, err := c.getVolume(ctx, key)
+	if err != nil {
+		return observeErr(err)
+	}
+
+	var current string
+
+	// If label already set, get value.
+	for k, v := range vol.Labels {
```
mhmxs, line 207: I guess `current := vol.Labels[ReservedLabelFailureMode]` does the same.
`var current string` is an empty string while `vol.Labels[ReservedLabelFailureMode]` returns value or empty string.

internal/pkg/storageos/volume_labels.go
```diff
@@ -162,3 +180,77 @@ func (c *Client) EnsureReplicas(ctx context.Context, key client.ObjectKey, desir
...
+		return observeErr(api.MapAPIError(err, resp))
+	}
+	return observeErr(nil)
+}
+
+// ParseFailureMode parses a string and returns either a failure mode intent or
+// a threshold, if set.  Only one of the intent or threshold should be set.
+func ParseFailureMode(s string) (api.FailureModeIntent, uint64, error) {
+	//  Return defaults if no value set.
+	if s == "" {
```
mhmxs, line 236: This expression should move int `switch` `case "", FailureModeHard`

internal/pkg/storageos/volume_labels.go
croomes: That's a good idea.  We use the current pattern in many places, so let's do a follow-up PR after `v2.4.0` to change everywhere.

internal/pkg/storageos/volume_labels.go
croomes: yes...  I do this often - I don't know why!

internal/pkg/storageos/volume_labels.go
croomes: yes, that's better.

mhmxs (approved):

PR #52, opened 2021-04-21
Introduce StorageClass to PVC annotation mutator

mhmxs:
This mutator looks for StorageClass of newly created PVC and sets an annotation with it's UID.

### DIFF
--- a/controllers/pvc-mutator/README.md
+++ b/controllers/pvc-mutator/README.md
@@ -12,6 +12,9 @@ task:
   ensures that PVCs that have requested encryption have a valid configuration,
   generating encryption keys if needed.
 
+- [StorageClass to annotation](/controllers/pvc-mutator/storageclass/README.md):
+  ensures that StorageOS related PVCs have them StorageClass as an annotation.
+
 ## Tunables
 
 There are currently no tunable flags for the PVC Mutator.
--- a/controllers/pvc-mutator/README.md
+++ b/controllers/pvc-mutator/README.md
@@ -13,7 +13,7 @@ task:
   generating encryption keys if needed.
 
 - [StorageClass to annotation](/controllers/pvc-mutator/storageclass/README.md):
-  ensures that StorageOS related PVCs have them StorageClass as an annotation.
+  ensures that StorageOS related PVCs have their StorageClass' UID as an annotation.
 
 ## Tunables
 
--- a/controllers/pvc-mutator/storageclass/README.md
+++ b/controllers/pvc-mutator/storageclass/README.md
@@ -1,12 +1,23 @@
-# PVC StorageClass mutator
+# PVC StorageClass Annotation Mutator
 
-The PVC StorageClass mutator tries to detect the StorageClass of the newly created PVC.
-It sets the `UID` of the StorageClass as an annotation on PVC.
+The PVC StorageClass Annotation Mutator tries to detect the StorageClass of the
+PVC being created and sets the StorageClass' UID in the
+`storageos.com/storageclass` annotation on the PVC.
+This allows StorageOS to detect whether the StorageClass parameters used to
+create the volume may have been changed later.  Since StorageClasses are
+immutable, deleting and recreating the StorageClass is the only way to change
+parameters.
 
 ## Trigger
 
 Only PVCs that will be provisioned by StorageOS are candidates for mutation.
 
+## Failure Policy
+
+Failure to set the StorageClass annotation should not cause the PVC creation to
+fail.
+
 ## Tunables
 
-There are currently no tunable flags for StorageClass Parameter Sync.
+There are currently no tunable flags for the PVC StorageClass Annotation
+Mutator.
--- a/controllers/pvc-mutator/storageclass/README.md
+++ b/controllers/pvc-mutator/storageclass/README.md
@@ -1,0 +1,12 @@
+# PVC StorageClass mutator
+
+The PVC StorageClass mutator tries to detect the StorageClass of the newly created PVC.
+It sets the `UID` of the StorageClass as an annotation on PVC.
+
+## Trigger
+
+Only PVCs that will be provisioned by StorageOS are candidates for mutation.
+
+## Tunables
+
+There are currently no tunable flags for StorageClass Parameter Sync.
--- a/controllers/pvc-mutator/storageclass/annotation.go
+++ b/controllers/pvc-mutator/storageclass/annotation.go
@@ -1,0 +1,62 @@
+package storageclass
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	"github.com/storageos/api-manager/internal/pkg/provisioner"
+	"github.com/storageos/api-manager/internal/pkg/storageos"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const AnnotationKey = storageos.ReservedLabelPrefix + "storageclass"
+
+// AnnotationSetter is responsible to add an annotation
+// to link the StorageClass with the PVC.
+type AnnotationSetter struct {
+	client.Client
+	log logr.Logger
+}
+
+// NewAnnotationSetter returns a new PVC annotation mutating admission
+// controller.
+func NewAnnotationSetter(k8s client.Client) *AnnotationSetter {
+	return &AnnotationSetter{
+		Client: k8s,
+		log:    ctrl.Log.WithName("storageclass"),
+	}
+}
+
+// MutatePVC mutates a given pvc with a new annotation,
+// by attached StorageClass.
+//
+// Errors returned here may block creation of the PVC, depending on the
+// FailurePolicy set in the webhook configuration.
+func (s *AnnotationSetter) MutatePVC(ctx context.Context, pvc *corev1.PersistentVolumeClaim, namespace string) error {
+	log := s.log.WithValues("pvc", client.ObjectKeyFromObject(pvc).String())
+	log.V(4).Info("received pvc for mutation")
+
+	// Find StorageClass of PVC
+	storageClass, err := provisioner.StorageClassForPVC(s.Client, pvc)
+	if err != nil {
+		return errors.Wrap(err, "failed to check pvc provisioner")
+	}
+
+	// Skip mutation if the PVC is not provisioned by StorageOS
+	if provisioned := provisioner.IsProvisionedStorageClass(storageClass, provisioner.DriverName); !provisioned {
+		log.V(4).Info("pvc was not provisioned by StorageOS, skipping")
+		return nil
+	}
+
+	// Set annotation on the PVC based on StorageClass
+	if pvc.Annotations == nil {
+		pvc.Annotations = make(map[string]string)
+	}
+	pvc.Annotations[AnnotationKey] = string(storageClass.UID)
+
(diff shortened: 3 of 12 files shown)


mhmxs:
More PVC test are coming, but somehow if i had added more it started to fail.

controllers/pvc-mutator/storageclass/annotation_test.go
```diff
@@ -0,0 +1,148 @@
...
+	testcases := []struct {
+		name         string
+		namespace    string
+		storageClass *storagev1.StorageClass
+		wantErr      bool
+	}{
+		{
+			name:      "empty StorageClass",
+			namespace: testNamespace,
+			wantErr:   true,
```
darkowlzz, line 64: Empty StorageClass shouldn't cause the mutation to return error. `StorageClassForPVC()` will return the default cluster StorageClass value if it's not set in the PVC.
For a failure case, setting the PVC StorageClass to a non-existent SC will work.

controllers/pvc-mutator/storageclass/annotation_test.go
```diff
@@ -0,0 +1,148 @@
...
+				if ok {
+					t.Error("annotation found for foreign")
+				}
+			}
+		})
+	}
+}
+
+// createPVC creates and returns a PVC object.
+func createPVC(name, namespace string, storageClass *storagev1.StorageClass) *corev1.PersistentVolumeClaim {
```
darkowlzz, line 125: Looks like we are creating `createPVC()` in every PVC mutator package.
Maybe separately, in the future, we should have a common test helper package that returns the core PVC instance given name, namespace and such package specific `createPVC()` functions can just wrap their requirements around the core share PVC creation helper function.

controllers/pvc-mutator/storageclass/annotation_test.go
```diff
@@ -0,0 +1,148 @@
...
+			AccessModes: []corev1.PersistentVolumeAccessMode{
+				corev1.ReadWriteOnce,
+			},
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: resource.MustParse("1Gi"),
+				},
+			},
+		},
+	}
```
darkowlzz, line 141: Looks like we are creating similar `createPVC()` in every PVC mutator package.
Maybe separately, in the future, we should have a common test helper package that returns the core PVC instance given name, namespace and such package specific `createPVC()` functions can just wrap their requirements around the core share PVC creation helper function.

controllers/pvc-mutator/storageclass/annotation.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Find StorageClass of PVC
+	storageClass, err := provisioner.StorageClassForPVC(s.Client, pvc)
+	if err != nil {
+		return errors.Wrap(err, "failed to check pvc provisioner")
+	}
+
+	// Skip mutation if the PVC is not provisioned by StorageOS
+	if provisioned := provisioner.IsProvisionedStorageClass(storageClass, provisioner.DriverName); !provisioned {
+		log.V(4).Info("pvc was not provisioned by StorageOS, skipping")
```
darkowlzz, line 50: Maybe `pvc does not provisioned by StorageOS` similar to the keygen package sounds better?

internal/pkg/provisioner/storageos.go
```diff
@@ -90,16 +91,22 @@ func IsProvisionedPVC(k8s client.Client, pvc corev1.PersistentVolumeClaim, names
...
+// IsProvisionedStorageClass returns true if the StorageClass was provided by one of
+// the given provisioners.
+func IsProvisionedStorageClass(sc *storagev1.StorageClass, provisioners ...string) bool {
 	// Check if the StorageClass provisioner matches with any of the provided
 	// provisioners.
 	for _, provisioner := range provisioners {
 		if sc.Provisioner == provisioner {
-			// This is a managed volume.
-			return true, nil
+			// This is a managed storage class.
```
darkowlzz, line 104: Seems a little incorrect. Volumes are managed by storageos if the provisioner value is storageos. But storageclasses are not managed by storageos. SC uses storageos provisioner.

internal/pkg/provisioner/storageos.go
```diff
@@ -90,16 +91,22 @@ func IsProvisionedPVC(k8s client.Client, pvc corev1.PersistentVolumeClaim, names
 		return false, err
 	}
 
+	return IsProvisionedStorageClass(sc, provisioners...), nil
+}
+
+// IsProvisionedStorageClass returns true if the StorageClass was provided by one of
+// the given provisioners.
```
darkowlzz, line 98: Maybe `if the StorageClass has one of the given provisioners`?

controllers/pvc-mutator/storageclass/annotation_test.go
mhmxs: I removed this created a separate case for the failure.

controllers/pvc_storageclass_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&pvc), &mutatedPVC)
+				if err != nil {
+					return ""
+				}
+
+				return mutatedPVC.Annotations[provisioner.StorageClassUUIDAnnotationKey]
+			}, timeout, interval).Should(Equal(string(persistedSC.UID)))
+		})
+	})
```
croomes, line 193: Even though the unit tests cover this, I'd really like there to be test cases that create PVCs with different StorageClass provisioners & defaults and checks the end result. The encryption mutator is a good example.

controllers/pvc-mutator/storageclass/README.md
```diff
@@ -0,0 +1,12 @@
+# PVC StorageClass mutator
```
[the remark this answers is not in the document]
croomes: Don't commit this directly - the patch won't apply cleanly.

controllers/pvc-mutator/storageclass/annotation.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Find StorageClass of PVC
+	storageClass, err := provisioner.StorageClassForPVC(s.Client, pvc)
+	if err != nil {
+		return errors.Wrap(err, "failed to check pvc provisioner")
+	}
+
+	// Skip mutation if the PVC is not provisioned by StorageOS
+	if provisioned := provisioner.IsProvisionedStorageClass(storageClass, provisioner.DriverName); !provisioned {
+		log.V(4).Info("pvc does not provisioned by StorageOS, skipping")
```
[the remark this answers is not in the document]
darkowlzz: Much better. Thanks.

controllers/pvc-mutator/README.md
```diff
@@ -12,6 +12,9 @@ task:
   ensures that PVCs that have requested encryption have a valid configuration,
   generating encryption keys if needed.
 
+- [StorageClass to annotation](/controllers/pvc-mutator/storageclass/README.md):
+  ensures that StorageOS related PVCs have them StorageClass as an annotation.
```
croomes, line 16:
```suggestion
  ensures that StorageOS related PVCs have their StorageClass' UID as an annotation.
```

croomes: nit: typo/grammar

croomes (approved):
Looks great!  Thanks for spending the extra time on all the feedback.

controllers/pvc-mutator/storageclass/annotation_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	namespace := "namespace"
+
+	// Create a pvc.
+	pvc := createPVC("pvc1", namespace, nil)
+
+	err := annotationSetter.MutatePVC(context.Background(), pvc, namespace)
+	if err == nil {
+		t.Fatal("this must fail")
```
darkowlzz, line 42: The message doesn't make it clear why the test failed. Like in the other tests, it'd be better to have some proper contextual info, as simple as "expected error, got none".

controllers/pvc_storageclass_test.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+
+		cancel()
+	})
+}
+
+var _ = Describe("PVC StorageClass UID to annotation controller", func() {
+	ctx := context.Background()
+
+	Context("When there is not StorageClass", func() {
```
darkowlzz, line 107:
```suggestion
	Context("When there is no StorageClass", func() {
```

controllers/pvc_storageclass_test.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+// Define utility constants for object names and testing timeouts and intervals.
+const (
+	storageClassTimeout  = time.Second * 1
+	storageClassInterval = time.Millisecond * 250
+)
+
+var (
+	defaultStorageClassName        = "stos-default"
```
darkowlzz, line 32: Let's change this name to reflect that it belongs to storageos. Maybe `stosDefaultStorageClassName`, similar to `foreignDefaultStorageClassName`. Found it confusing while reading the test below.

controllers/pvc_storageclass_test.go
```diff
@@ -0,0 +1,320 @@
...
+	givenStorageClassName          = "stos"
```
darkowlzz, line 33: Similar to the default SC name, this can be renamed to `stosStorageClassName` for clarity.

controllers/pvc_storageclass_test.go
```diff
@@ -0,0 +1,320 @@
...
+		It("The pvc should not be created", func() {
+			By("Expecting the PVC has default StorageClass")
+			pvc := genPVC()
+
+			By("Creating the PVC")
+			Expect(k8sClient.Create(ctx, &pvc)).ShouldNot(Succeed())
+		})
+	})
+
+	Context("When the default is not StorageOS not given", func() {
```
darkowlzz, line 119: Maybe `When the default StorageClass is non-StorageOS StorageClass`

controllers/pvc_storageclass_test.go
```diff
@@ -0,0 +1,320 @@
...
+				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&pvc), &got)
+				if err != nil {
+					return corev1.PersistentVolumeClaim{}
+				}
+				return got
+			}, storageClassTimeout, storageClassInterval).Should(Equal(pvc))
+		})
+	})
+
+	Context("When the default is not StorageOS given is not StorageOS", func() {
```
darkowlzz, line 145: Since this and the other context messages in the test case above and below seem to be a bit complex with two variations that can form a test matrix:
- Default StorageClass
- PVC's StorageClass

Maybe the tests would be simpler if we have two `Describe` blocks for when the default SC is storageos SC and non-storageos SC. And within them, we can have cases for PVCs that use different SCs.

I hope that'll simplify the Describe/Context messages for each of the tests. In the above Context, I would try to change it to `When the default SC is not STOS SC and the given PVC uses non-STOS SC`. It's a really long description of the test.

mhmxs: Originally i did those two Context-es and created multiple It-s, but i had issues with context.Contex and also other resource creation in BeforeEach. I thought I'm totally wrong and instead of refactoring the entire setup which now looks similar than keygen, i went on this way. But creating a second Describe makes much more sense. Because I also feel nausea if that long names ;) Thanks

darkowlzz (approved):
The tests look really good now. Thanks.
Please squash and merge the changes.

PR #53, opened 2021-04-21
Add defaults from StorageClass

croomes:
Prior to applying labels after a change has been detected, defaults are loaded from the StorageClass parameters that provisioned the volume.  This ensures that the defaults set in the StorageClass are not overwritten because there is not a corresponding label on the PVC.

Since StorageClasses are immutable the params can't be changed after the volume has been provisioned.  To protect against the SC being deleted and re-added with different params, a reference to the original SC's UID is stored in an annotation on the PVC.  If the current SC does not match, an error is logged and label sync is disabled for that PVC.


mhmxs (approved):

darkowlzz (approved):
LGTM!

PR #55, opened 2021-04-23
Use composite client in admission controllers

darkowlzz:
This change replaces all the cached/uncached clients in admission
controllers with a composite client. The composite client will help
handle the cases where the cached client can't find the recently
created resources. The composite client will retry the query using
the raw client to query the k8s API server directly.

Refer https://pkg.go.dev/github.com/darkowlzz/operator-toolkit@v0.0.0-20210423182126-723c7a1e8b45/client/composite for details about the composite client.

### DIFF
--- a/vendor/github.com/darkowlzz/operator-toolkit/client/composite/client.go
+++ b/vendor/github.com/darkowlzz/operator-toolkit/client/composite/client.go
@@ -1,0 +1,68 @@
+package composite
+
+import (
+	"context"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// Client is a composite client, composed of cached and uncached clients. It
+// can be used to query objects from cache and fall back to use the raw client
+// and get the object directly from the API server when there's a cache miss.
+type Client struct {
+	client.Client
+	Options
+
+	uncached client.Client
+}
+
+// ClientOption is used to configure the client.
+type Options struct {
+	// RawListing is used to perform raw listing operations, uncached.
+	RawListing bool
+}
+
+// NewClient creates and returns a composite Client.
+func NewClient(cached client.Client, uncached client.Client, opts Options) *Client {
+	return &Client{
+		Client:   cached,
+		Options:  opts,
+		uncached: uncached,
+	}
+}
+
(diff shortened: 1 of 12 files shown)


vendor/github.com/darkowlzz/operator-toolkit/client/composite/client.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// Client is a composite client, composed of cached and unclient clients. It
```
croomes, line 11: For later:
unclient -> uncached

vendor/github.com/darkowlzz/operator-toolkit/client/composite/client.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, err
+	}
+
+	return NewClient(mgr.GetClient(), uncachedClient, opts), nil
+}
+
+// Get first fetches the object using the cached client. If the object is not
+// found in the cached client, it retries using the uncached client.
+func (c *Client) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+	if cErr := c.Client.Get(ctx, key, obj); cErr != nil {
```
croomes, line 51: q: why `cErr` and not just `err`?

croomes (approved):
Looks good, thanks!

I re-ran the `TestPVCLabelsUpdate` kubecover test that was failing intermittently with the StorageClass not found error and couldn't reproduce after about 10 tries.  I'll re-run the full suite & keep testing, but suggest you merge.

mhmxs (approved):

vendor/github.com/darkowlzz/operator-toolkit/client/composite/client.go
darkowlzz: I think there were some other error variables before that I removed.
This is part of a change in style I've been practicing to avoid shadowing variables accidentally, especially error variables.

vendor/github.com/darkowlzz/operator-toolkit/client/composite/client.go
darkowlzz: thanks.

PR #56, opened 2021-04-23
Test flake: Random endpoint ip address should not be a loopback address

croomes:
Fixes an e2e test flake where if the randomly generated endpoint IP address is a loopback address then the Kubernetes API will reject it.  Validation function: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L5769-L5789

Test failure https://github.com/storageos/api-manager/pull/55/checks?check_run_id=2422375036:
```
failed to create endpoints resource: Endpoints \"pvc-d6941232-6432-43bf-b5de-59c999641a09\" is invalid: subsets[0].addresses[0].ip: Invalid value: \"127.136.127.136\": may not be in the loopback range (127.0.0.0/8)"
```

The crude fix is to ensure the first octet is between 1-101.

### DIFF
--- a/internal/pkg/storageos/mock_client.go
+++ b/internal/pkg/storageos/mock_client.go
@@ -434,7 +434,7 @@ func (c *MockClient) RandomVol() *SharedVolume {
 		ServiceName:      "pvc-" + uuid.New().String(),
 		PVCName:          randomString(8),
 		Namespace:        "default",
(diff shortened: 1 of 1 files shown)


mhmxs (approved):

darkowlzz (approved):
Nice. Thanks.

PR #57, opened 2021-04-27
Fix storageclass annotation int. tests

mhmxs:
Previously the test compared the entire PVC with itself on case of "changes not needed" which doesn't make much sense and because there was a misconfiguration in test cases this was hidden. I fixed the config and decrease the scope of comparison.

### DIFF
--- a/controllers/pvc_storageclass_test.go
+++ b/controllers/pvc_storageclass_test.go
@@ -139,7 +139,7 @@ var _ = Describe("The default StorageClass has not been configured", func() {
 				Expect(k8sClient.Delete(ctx, &pvc)).Should(Succeed())
 			}()
 
-			By("Expecting the PVC hasn't annotated with StorageClass")
+			By("Expecting the PVC not to have the StorageClass annotation")
 			// The create operation mutates the PVC, so the testcase has
 			// reference only to the final version. Only annotations could be tested.
 			Consistently(func() string {
@@ -174,7 +174,7 @@ var _ = Describe("The default StorageClass has not been configured", func() {
 			err := k8sClient.Get(ctx, client.ObjectKey{Name: storageClass.Name}, &persistedSC)
 			Expect(err).NotTo(HaveOccurred(), "failed to fetch StorageClass")
 
-			By("Expecting the PVC has annotated with StorageClass")
+			By("Expecting the PVC has the StorageClass annotation")
 			// The create operation mutates the PVC, so the testcase has
 			// reference only to the final version. Only annotations could be tested.
 			Eventually(func() string {
@@ -208,7 +208,7 @@ var _ = Describe("The default StorageClass is not StorageOS", func() {
 				Expect(k8sClient.Delete(ctx, &pvc)).Should(Succeed())
 			}()
 
-			By("Expecting the PVC hasn't annotated with StorageClass")
+			By("Expecting the PVC not to have the StorageClass annotation")
 			// The create operation mutates the PVC, so the testcase has
 			// reference only to the final version. Only annotations could be tested.
 			Consistently(func() string {
@@ -239,7 +239,7 @@ var _ = Describe("The default StorageClass is not StorageOS", func() {
 				Expect(k8sClient.Delete(ctx, &pvc)).Should(Succeed())
 			}()
 
-			By("Expecting the PVC hasn't annotated with StorageClass")
+			By("Expecting the PVC not to have the StorageClass annotation")
 			// The create operation mutates the PVC, so the testcase has
 			// reference only to the final version. Only annotations could be tested.
 			Consistently(func() string {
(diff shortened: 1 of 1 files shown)


controllers/pvc_storageclass_test.go
```diff
@@ -331,20 +335,21 @@ var _ = Describe("The default StorageClass is StorageOS", func() {
 			}()
 
 			By("Expecting the PVC has to be unchanged")
-			Consistently(func() corev1.PersistentVolumeClaim {
+			Consistently(func() string {
 				got := corev1.PersistentVolumeClaim{}
 				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&pvc), &got)
 				if err != nil {
-					return corev1.PersistentVolumeClaim{}
+					return "kube get error"
```
darkowlzz, line 342: Maybe just `err.Error()` ? Rather than a custom string. Doesn't matters much.

controllers/pvc_storageclass_test.go
```diff
@@ -315,8 +319,8 @@ var _ = Describe("The default StorageClass is StorageOS", func() {
 	})
 
 	Context("When the given StorageClass is not StorageOS", func() {
-		defaultStorageClass := getStorageClass(stosDefaultStorageClassName, true)
-		storageClass := getStorageClass(nonStosStorageClassName, false)
+		defaultStorageClass := getStorageClass(stosDefaultStorageClassName, provisioner.DriverName, true)
+		storageClass := getStorageClass(nonStosStorageClassName, "foo-provisioner", false)
```
darkowlzz, line 323: We can use the new `nonStosProvisioner` constant here 🙂

controllers/pvc_storageclass_test.go
```diff
@@ -231,20 +234,21 @@ var _ = Describe("The default StorageClass is not StorageOS", func() {
 			}()
 
 			By("Expecting the PVC has to be unchanged")
-			Consistently(func() corev1.PersistentVolumeClaim {
+			Consistently(func() string {
 				got := corev1.PersistentVolumeClaim{}
 				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&pvc), &got)
 				if err != nil {
-					return corev1.PersistentVolumeClaim{}
+					return "kube get error"
```
darkowlzz, line 241: Maybe `err.Error()`?

controllers/pvc_storageclass_test.go
```diff
@@ -139,19 +140,20 @@ var _ = Describe("The default StorageClass has not been configured", func() {
...
 				}
-				return got
-			}, storageClassTimeout, storageClassInterval).Should(Equal(pvc))
+
+				return got.Annotations[provisioner.StorageClassUUIDAnnotationKey]
+			}, storageClassTimeout, storageClassInterval).Should(BeEmpty())
```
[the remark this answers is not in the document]
darkowlzz: Ensuring we don't change anything else in the PVC sounds better since we are taking the whole object and changing only a part of it.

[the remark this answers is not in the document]
croomes: I spoke with @mhmxs about this and he explained why we can't do this - the pre-create PVC won't have metadata like the PVC UID so a compare of the whole object will always fail.  It passed before as I missed that because I was passing in pvc as a reference, the create call was updating it with the post-create object.  The test always passes.

So I think it's ok to only check that we don't add the annotation.  But the test case should be renamed to explain this, perhaps with a comment why we don't compare the whole object.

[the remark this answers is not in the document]
mhmxs: Thank you for your describe. I haven't commented the code, because other scenarios also test annotation only. I left some comment.

controllers/pvc_storageclass_test.go
```diff
@@ -138,20 +139,23 @@ var _ = Describe("The default StorageClass has not been configured", func() {
 				Expect(k8sClient.Delete(ctx, &pvc)).Should(Succeed())
 			}()
 
-			By("Expecting the PVC has to be unchanged")
-			Consistently(func() corev1.PersistentVolumeClaim {
+			By("Expecting the PVC hasn't annotated with StorageClass")
```
croomes, line 142: sorry, just a nit but it's bugging me:
`Expecting the PVC hasn't annotated with StorageClass` -> `Expecting the PVC not to have the StorageClass annotation`.

controllers/pvc_storageclass_test.go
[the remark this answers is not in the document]
croomes: thanks!

controllers/pvc_storageclass_test.go
```diff
@@ -263,7 +275,9 @@ var _ = Describe("The default StorageClass is not StorageOS", func() {
 			err := k8sClient.Get(ctx, client.ObjectKey{Name: storageClass.Name}, &persistedSC)
 			Expect(err).NotTo(HaveOccurred(), "failed to fetch StorageClass")
 
-			By("Expecting the PVC to be patched")
+			By("Expecting the PVC has annotated with StorageClass")
```
croomes, line 278: Again, just a nit, but this would read better: `Expecting the PVC has the StorageClass annotation`

croomes (approved):
Looks good - just those two grammar changes.

PR #58, opened 2021-04-29
Verify annotations in keygen test

mhmxs:
"the pre-create PVC won't have metadata like the PVC UID so a compare of the whole object will always fail. It passed before as I missed that because I was passing in pvc as a reference, the create call was updating it with the post-create object. The test always passes." Thanks @croomes


croomes (approved):
Thanks!

PR #61, opened 2021-05-03
Check encryption flag in storage class

mhmxs:
Original version checked encryption label only at the PVC, but encryption should be inherited from storage class. This change adds this ability.

The original encryption label parser forced un-encrypted storage creation if the label was invalid. This behavior leads to non encrypted volumes if user made a typo in label, which is not the safest way to handle the case. After this change invalid label value will block the PVC creation.

### DIFF
--- a/controllers/pvc-mutator/encryption/keygen.go
+++ b/controllers/pvc-mutator/encryption/keygen.go
@@ -101,21 +101,29 @@ func (s *EncryptionKeySetter) MutatePVC(ctx context.Context, pvc *corev1.Persist
 	log := s.log.WithValues("pvc", client.ObjectKeyFromObject(pvc).String())
 	log.V(4).Info("received pvc for mutation")
 
-	// Skip mutation if the PVC is not provisioned by StorageOS
-	provisioned, err := provisioner.IsProvisionedPVC(s.Client, *pvc, namespace, provisioner.DriverName)
+	// Retrieve StorageClass of the PVC.
+	storageClass, err := provisioner.StorageClassForPVC(s.Client, pvc)
 	if err != nil {
-		return errors.Wrap(err, "failed to check pvc provisioner")
+		return errors.Wrap(err, "failed to retrieve storageclass of pvc")
 	}
+
+	// Skip mutation if the PVC is not provisioned by StorageOS
+	provisioned := provisioner.IsProvisionedStorageClass(storageClass, provisioner.DriverName)
 	if !provisioned {
 		log.V(4).Info("pvc does not provisioned by StorageOS, skipping")
 		return nil
 	}
 
-	// Skip mutation if the PVC does not have encryption enabled.  Don't bother checking
-	// the StorageClass to make sure it's StorageOS.  The encryption label
-	// should only be added to StorageOS PVCs.
-	if !isEnabled(s.enabledLabel, pvc.GetLabels()) {
-		log.V(4).Info(fmt.Sprintf("pvc does not have %s=true annotation, skipping", s.enabledLabel))
+	// Skip mutation if the PVC does not have encryption enabled.
+	// The encryption label should be added to StorageOS PVCs
+	// or inherited from StorageOS StorageClass.
+	// Invalid value of encryption must block PVC creation.
+	enabled, err := s.isEnabled(pvc.GetLabels(), storageClass.Parameters)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to parse boolean value of %s pvc", s.enabledLabel))
+	}
+	if !enabled {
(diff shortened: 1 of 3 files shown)


controllers/pvc-mutator/encryption/keygen.go
```diff
@@ -101,21 +101,29 @@ func (s *EncryptionKeySetter) MutatePVC(ctx context.Context, pvc *corev1.Persist
...
-	// should only be added to StorageOS PVCs.
-	if !isEnabled(s.enabledLabel, pvc.GetLabels()) {
-		log.V(4).Info(fmt.Sprintf("pvc does not have %s=true annotation, skipping", s.enabledLabel))
+	// Skip mutation if the PVC does not have encryption enabled.
+	// The encryption label should be added to StorageOS PVCs
+	// or inherited from StorageOS StorageClass.
+	// Invalid value of encryption must block PVC creation.
+	enabled, err := s.isEnabled(pvc.GetLabels(), storageClass.Parameters)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to parse boolean value of %s pvc", s.enabledLabel))
```
darkowlzz, line 123: This would render something like:
```
failed to parse boolean value of storageos/encryption pvc
```

Maybe the message could be improved?
Also, I'm not sure if as a caller of `isEnabled()`, the error here should mention about boolean parsing. The caller need not be specific about how the result is determined in that function. Just a thought.

controllers/pvc-mutator/encryption/keygen.go
```diff
@@ -212,16 +220,14 @@ func GenerateVolumeSecretName() string {
 	return fmt.Sprintf("%s-%s", VolumeSecretNamePrefix, uuid.New().String())
 }
 
-// isEnabled returns true if the key is set in the kv map and its value is true.
-func isEnabled(key string, kv map[string]string) bool {
-	val, exists := kv[key]
-	if !exists {
-		return false
+// isEnabled iterates on given maps and looks for encryption key. First occurence wins.
```
darkowlzz, line 223:
```suggestion
// isEnabled iterates on given maps and looks for encryption key. First occurrence wins.
```

controllers/pvc-mutator/encryption/keygen.go
```diff
@@ -101,21 +101,29 @@ func (s *EncryptionKeySetter) MutatePVC(ctx context.Context, pvc *corev1.Persist
...
+	storageClass, err := provisioner.StorageClassForPVC(s.Client, pvc)
 	if err != nil {
-		return errors.Wrap(err, "failed to check pvc provisioner")
+		return errors.Wrap(err, "failed to retrieve storageclass of pvc")
 	}
+
+	// Skip mutation if the PVC is not provisioned by StorageOS
+	provisioned := provisioner.IsProvisionedStorageClass(storageClass, provisioner.DriverName)
 	if !provisioned {
 		log.V(4).Info("pvc does not provisioned by StorageOS, skipping")
```
croomes, line 113: Can you change this while you're here:

`pvc does not provisioned by StorageOS, skipping` -> `pvc will not be provisioned by StorageOS, skipping`

controllers/pvc-mutator/encryption/keygen.go
mhmxs: But somehow the caller has to realize the error, because forcing a value on case of parse error should lead different behavior than the user expected. Plain volume instead of encrypted or the opposite.

croomes (approved):
Looks good - thanks.  Good to merge after you've addressed the review comments.

controllers/pvc-mutator/encryption/keygen.go
darkowlzz: Maybe just `failed to parse "storageos/encryption" option on PVC` should be good. The wrapped error will have the details.

darkowlzz (approved):
Nice fix and tests. LGTM!

controllers/pvc-mutator/encryption/keygen.go
mhmxs: `failed to parse "storageos/encryption" option for PVC` because we parse both PVC and storageclass?

darkowlzz: oh right! Maybe volume instead of PVC would be better?

darkowlzz: oh I see. `for`. Sounds right to me.

croomes: How about: `failed to parse boolean value for "storageos.com/encryption" pvc label or storageclass parameter`?  I think that covers both, and explains what's expected.

PR #62, opened 2021-05-06
Don't attempt service creation if the owning PVC doesn't exist

croomes:
Fixes a log error where if the PVC wasn't found, the NFS service creation would still be attempted and fail since the OwnerReference UID would be empty.  Now it doesn't attempt the service creation until it can set the owner properly.

This removes a scary log message only - it doesn't change the behaviour.  For as long as the StorageOS volume is attached as RWX and there is no corresponding k8s service, the service creation will be retried.

### DIFF
--- a/internal/controllers/sharedvolume/controller.go
+++ b/internal/controllers/sharedvolume/controller.go
@@ -167,11 +167,9 @@ func (r *Reconciler) Start(ctx context.Context) error {
 			// is no longer required and we can ignore the request.
 			pvc := &corev1.PersistentVolumeClaim{}
 			if err := r.Client.Get(ctx, types.NamespacedName{Name: vol.PVCName, Namespace: vol.Namespace}, pvc); err != nil {
-				if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


mhmxs (approved):

darkowlzz (commented):
There seems to be some issue with the `wantErr` check.
If you run the test, you'd see log message like

```
    controller_test.go:369: SharedVolume.Reconcile() error = <nil>, wantErr true
```

But it's not causing the test to fail. Setting `wantErr` has no effect. It's just a log https://github.com/storageos/api-manager/blob/11af62270d82477835da41958801f6fc735c4c9d/internal/controllers/sharedvolume/controller_test.go#L369.
The target error for this change is the PVC Get error, but since it's not returned by the for loop in the reconciler, it can't be tested. In the current state, even if there's a failure, the test runs for 5 seconds, waiting for nothing.
Maybe it'd be better to move the core for loop into an independent function to make it more testable. Can be done separately.

avestuk (approved):

mhmxs (approved):

PR #63, opened 2021-05-13
Compare SC and PVC creation time during label sync

mhmxs:
Currently if label sync finds the UID annotation it compares it to the current version of the StorageClass.
If it doesn't match, stop syncing the labels and some user action is needed.
If it isn't found it fetches the StorageClass and passes the parameters as labels.
In some edge cases we have the risk that we synchronize wrong parameters, because the parameters were different in the time of the PVC creation.
This change introduces usage of creation time to more safely use the underlying StorageClass, if SC is older, use it, if younger just stop syncing and ask the user to do some action.

### DIFF
--- a/internal/pkg/provisioner/storageclass.go
+++ b/internal/pkg/provisioner/storageclass.go
@@ -99,10 +99,11 @@ func StorageClassReservedParams(sc *storagev1.StorageClass) map[string]string {
 }
 
 // ValidateOrSetStorageClassUID returns true if the StorageClass annotation on
-// the PVC object matches the uid passed in or the StorageClass is younger than PVC.
+// the PVC object matches the UID of the passed in StorageClass
+// or the StorageClass has created before the given PVC.
 //
-// If the annotation does not exist and the StorageClass is younger than PVC,
-// it sets the uid passed in as the new StorageClass annotation.
+// If the annotation does not exist and the StorageClass has created before the PVC,
+// it sets the UID of the passed in StorageClass as the new StorageClass annotation.
 func ValidateOrSetStorageClassUID(ctx context.Context, k8s client.Client, sc client.Object, pvc client.Object) (bool, error) {
 	provisionedUID := pvc.GetAnnotations()[StorageClassUUIDAnnotationKey]
 	if provisionedUID != "" {
@@ -110,7 +111,7 @@ func ValidateOrSetStorageClassUID(ctx context.Context, k8s client.Client, sc cli
 		return string(sc.GetUID()) == provisionedUID, nil
 	}
 
-	// If StorageClass is younger than the PVC itself, the synchronization isn't safe.
+	// If StorageClass has created after the PVC itself, the synchronization isn't safe.
 	if sc.GetCreationTimestamp().Unix() > pvc.GetCreationTimestamp().Unix() {
 		return false, nil
 	}
--- a/internal/pkg/provisioner/storageclass.go
+++ b/internal/pkg/provisioner/storageclass.go
@@ -99,29 +99,33 @@ func StorageClassReservedParams(sc *storagev1.StorageClass) map[string]string {
 }
 
 // ValidateOrSetStorageClassUID returns true if the StorageClass annotation on
-// the PVC object matches the uid passed in.
+// the PVC object matches the uid passed in or the StorageClass is younger than PVC.
 //
-// If the annotation does not exist, it sets the uid passed in as the new
-// StorageClass annotation.
-func ValidateOrSetStorageClassUID(ctx context.Context, k8s client.Client, uid types.UID, obj client.Object) (bool, error) {
-	provisionedUID := obj.GetAnnotations()[StorageClassUUIDAnnotationKey]
+// If the annotation does not exist and the StorageClass is younger than PVC,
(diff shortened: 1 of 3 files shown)


internal/pkg/provisioner/storageclass.go
```diff
@@ -99,29 +99,33 @@ func StorageClassReservedParams(sc *storagev1.StorageClass) map[string]string {
...
-// the PVC object matches the uid passed in.
+// the PVC object matches the uid passed in or the StorageClass is younger than PVC.
 //
-// If the annotation does not exist, it sets the uid passed in as the new
-// StorageClass annotation.
-func ValidateOrSetStorageClassUID(ctx context.Context, k8s client.Client, uid types.UID, obj client.Object) (bool, error) {
-	provisionedUID := obj.GetAnnotations()[StorageClassUUIDAnnotationKey]
+// If the annotation does not exist and the StorageClass is younger than PVC,
+// it sets the uid passed in as the new StorageClass annotation.
+func ValidateOrSetStorageClassUID(ctx context.Context, k8s client.Client, sc client.Object, pvc client.Object) (bool, error) {
```
croomes, line 106: nit: I'd keep the `client.Object` var name as `obj` instead of `pvc` as it makes it more clear what type it is.  Below where you use its key to fetch the PVC can it `pvc` instead of `v1Pvc`.

croomes (approved):
LGTM - thanks

internal/pkg/provisioner/storageclass.go
mhmxs: We have two `client.Object`s at the moment, so somehow i wanted to inform clients which one is which.

croomes: ah, I see. ok

internal/pkg/provisioner/storageclass.go
```diff
@@ -99,29 +99,33 @@ func StorageClassReservedParams(sc *storagev1.StorageClass) map[string]string {
...
 // ValidateOrSetStorageClassUID returns true if the StorageClass annotation on
-// the PVC object matches the uid passed in.
+// the PVC object matches the uid passed in or the StorageClass is younger than PVC.
 //
-// If the annotation does not exist, it sets the uid passed in as the new
-// StorageClass annotation.
-func ValidateOrSetStorageClassUID(ctx context.Context, k8s client.Client, uid types.UID, obj client.Object) (bool, error) {
-	provisionedUID := obj.GetAnnotations()[StorageClassUUIDAnnotationKey]
+// If the annotation does not exist and the StorageClass is younger than PVC,
+// it sets the uid passed in as the new StorageClass annotation.
```
darkowlzz, line 105: Since the uid is no longer being passed in but a SC, this should be updated to `sets the uid of the StorageClass as the new...`

internal/pkg/provisioner/storageclass.go
```diff
@@ -99,29 +99,33 @@ func StorageClassReservedParams(sc *storagev1.StorageClass) map[string]string {
 }
 
 // ValidateOrSetStorageClassUID returns true if the StorageClass annotation on
-// the PVC object matches the uid passed in.
+// the PVC object matches the uid passed in or the StorageClass is younger than PVC.
```
darkowlzz, line 102: This too can be improved based on the change. `matches the uid of the passed in SC...`.
And should it return true if the SC is older than PVC? As per the comments below, younger SC isn't safe and returns false.

darkowlzz (approved):
Implementation and tests LGTM.
Left a few comments about the comment lines change may require updating with this change.

internal/pkg/provisioner/storageclass.go
mhmxs: Sorry mate, i don't understand the second part. It returns true if the SC is older and updates was ok: https://github.com/storageos/api-manager/blob/5dd47fbacd5ece908761c3f6612226df916ed7d5/internal/pkg/provisioner/storageclass.go#L130

Could you please explain?

darkowlzz: Yes, so it should return true if SC is older. The above comment states that it is returning true if SC is younger.

Referring to https://github.com/storageos/api-manager/pull/63/files#diff-9c56144171130b386f3a5dab17f4c3af7fc963fdda728ce78287cb89ab31d1dcR113-R116

> 	// If StorageClass is younger than the PVC itself, the synchronization isn't safe.

internal/pkg/provisioner/storageclass.go
darkowlzz: Revisiting the comment above, line 104, it again says the opposite. If SC is younger than PVC, false is returned. uid isn't set.

croomes: agreed, it should be:
```
// If the annotation does not exist and the StorageClass is older than the PVC,
// it sets the uid passed in as the new StorageClass annotation.
```

croomes: personally I find the younger/older a bit confusing to read, but it's just me.  Maybe "was created [before|after]" would be better?

darkowlzz: Yes, I felt the same. Before and after sound better.

PR #71, opened 2021-07-29
Add Volume API and update Node API with CRDs

darkowlzz:
- Add the Volume API based on previous work by @croomes .
- Make both the Volume and Node CRDs cluster scoped.
- The Node API is moved to a new group called `api`, making the APIVersion to be `api.storageos.com/v1`. This is to avoid no group API issues that we have with old operator CRDs. The Volume CRD also uses the `api` group.
- Update CRDs for both Node and Volume API definitions.

These Node and Volume CRDs can be registered with k8s-apiserver to store Node and Volume objects.


mhmxs (approved):

croomes (approved):
LGTM - thanks!

PR #72, opened 2021-08-03
Restricted labels must be sent in node label update

croomes:
Fixes the error: `"msg":"failed to resync metadata to external object","reconciler":"node-label-sync","name":"nodename","error":"reserved label \"storageos.com/computeonly\" cannot be modified directly"}`.

An incorrect set of labels was being passed to the update call, causing the existing reserved label `storageos.com/computeonly` not to be sent in the updated request, triggering a validation error in the API.  The API requires that the existing reserved labels are included in the update request, without their values changing.


mhmxs (approved):
Lgtm

darkowlzz (approved):
LGTM!
Thanks.

PR #73, opened 2021-08-04
Add User, GroupPolicy and Cluster CRDs

darkowlzz:
User, GroupPolicy and Cluster CRDs are cluster scoped APIs.


croomes (dismissed):
Thanks, LGTM

api/v1/cluster_types.go
```diff
@@ -0,0 +1,78 @@
...
+// ClusterStatus defines the observed state of Cluster.
+type ClusterStatus struct{}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Cluster
+
+// Cluster is the Schema for the clusters API.
+type Cluster struct {
+	metav1.TypeMeta   `json:",inline"`
```
mhmxs, line 60: Do we need the comma?

api/v1/user_types.go
```diff
@@ -0,0 +1,69 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UserSpec defines the desired state of User.
+type UserSpec struct {
+	// SecretRefName is the name of a Secret that contains the user
+	// credentials.
+	SecretRefName string `json:"secretRefName,omitempty"`
+	// IsAdmin is used to set a user as an admin.
+	IsAdmin bool `json:"isAdmin,omitempty"`
```
mhmxs, line 37: Does it make sense to create a separated `Role` type where admin is only one role? This looks not so flexible.

mhmxs (dismissed):

api/v1/cluster_types.go
darkowlzz: Yes, that's not the json field name and comma is needed to separate the name and option.
This is a generated code, btw.

api/v1/user_types.go
darkowlzz: I'm not sure about that. I don't think there are any other roles other than admin and non-admin. There's group with policies to create restricted groups, but that's different.

PR #75, opened 2021-08-16
Sync StorageOS volumes to K8s api server

mhmxs:


internal/controllers/sharedvolume/controller.go
```diff
@@ -54,177 +61,149 @@ type Reconciler struct {
...
-					span.End()
-					continue
-				}
-				if cachedVol.IsEqual(vol) {
-					span.AddEvent("no update needed")
-					span.End()
-					continue
-				}
-			}
+	sv := storageos.SharedVolume{Volume: vol}
```
croomes, line 144: I'm wondering if you should use a constructor here - to be a valid SharedVolume it needs a few things (nfs endpoint, pvcName, etc.)  It might be better to validate in the constructor and return an error early?

croomes (commented):
Looks really good - I've tested and it works well.  Just some general comments/questions.

mhmxs:
Depends on: https://github.com/storageos/operator/pull/5

internal/controllers/sync-volume/controller.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+		span.SetAttributes(label.Int("volumes", len(volumes)))
+
+		// Current volumes for easy lookup later.
+		currentVolumes := map[string]client.ObjectKey{}
+
+		for _, obj := range volumes {
+			// Skip non-k8s volumes.  The PV name & PVC namespace will be used as the
+			// Service and Endpoints name & namespace.  The PVC name is required to set
+			// the Service ownerRef to the PVC.
```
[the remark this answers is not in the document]
croomes: As discussed, let's create objects for all volumes.  Volumes created by the CLI may not work in k8s though.

internal/controllers/sharedvolume/controller.go
croomes: For SharedVolumes, we require certain labels to be set.  The constructor should validate and return an error if the volume object isn't compatible.

mhmxs:
@croomes I have rebased this PR to the latest.

PR #86, opened 2021-09-01
Sync nodes from StorageOS

mhmxs:
This change moves StorageOS node API poll into a separated controller. After each poll the controller compares node content with last known content and persists node object if needed as custom resource.


mhmxs:
Tracked here: https://github.com/storageos/api-manager/pull/98

PR #93, opened 2021-09-17
Generate working version of manifests

mhmxs:
This change recreates manifests based on the working version in operator.


Makefile
```diff
@@ -44,7 +44,7 @@ help: ## Display this help.
 ##@ Development
 
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName="storageos-api-manager" webhook paths="./..." output:crd:artifacts:config=config/crd/bases output:webhook:artifacts:config=config/webhook/bases
```
croomes, line 47: Not sure it matters here, but the operator will name the role `storageos:api-manager`.

config/default/kustomization.yaml
```diff
@@ -1,36 +1,41 @@
...
 # field above.
-namePrefix: api-manager-
+# namePrefix: storageos-
 
 # Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+commonLabels:
+  app: storageos
+  app.kubernetes.io/component: storageos-api-manager
```
croomes, line 14: Just `api-manager`?  In other places, we drop the `storageos-` prefix for the component label.

config/rbac/key-management-role-binding.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: storageos-key-management
```
croomes, line 4: `storageos:key-management`

mhmxs: Controller runtime doesn't support names with `:` that's why i changed every to `-`. The error during generation is:
```
Error: unable to parse option "rbac:roleName=storageos:api-manager": [expected comma, got ":" (at <input>:1:19) extra arguments provided: "api-manager" (at <input>:1:20)]
```

Makefile
mhmxs: Answer same as : https://github.com/storageos/api-manager/pull/93#discussion_r711997010

croomes (approved):
LGTM after pipeline passes - thanks!

PR #106, opened 2021-12-10
add master node id to volume CR labels

Mojachieee:
This enables us to use a label selector to watch Volume CRs by node id, which will make our upgrade-guard more efficient


internal/controllers/sync-volume/utils.go
```diff
@@ -64,6 +64,12 @@ func getConverter(vol *stosv1.Volume, obj storageos.Object) func() error {
 			replicas = &r
 		}
 
+		// This enables us to use a label selector on the node ID
+		if vol.ObjectMeta.Labels == nil {
+			vol.ObjectMeta.Labels = make(map[string]string)
+		}
+		vol.ObjectMeta.Labels["master_node_id"] = stosVol.Master.NodeID
```
mhmxs, line 71: We usually use `-` instead of `_` for separator. Could you please change it?

mhmxs (approved):

PR #108, opened 2022-01-04
use replica count from label as spec instead of status value

Mojachieee:
This change ensures we set the desired replica count in the spec value of our CRD. 

Before this change we were setting the actual replica count to the spec value, which can cause confusion as this values may not match 

If the volume has any replicas the replica label will be there. The ondat controlplane will always set it


internal/controllers/sync-volume/utils.go
```diff
@@ -59,9 +61,15 @@ func getConverter(vol *stosv1.Volume, obj storageos.Object) func() error {
 		}
 
 		var replicas *uint8
-		if stosVol.Replicas != nil {
-			r := uint8(len(*stosVol.Replicas))
-			replicas = &r
+		replicaCount, ok := stosVol.Labels[storageos.ReservedLabelReplicas]
+		if ok {
+			count, err := strconv.Atoi(replicaCount)
+			if err == nil {
```
mhmxs, line 67: Could you please at least log if conversion fails? For me this is a bit optimistic solution :)

szank (approved):

mhmxs (approved):

PR #110, opened 2022-01-11
remove uint64 types from crd, typecast results from internal api

Mojachieee:
This will stop the crashes caused by trying to DeepCopy uint64 types. 

uint64 types are still used in the control plane, but we will now at least have logs when we fail to type cast them


internal/controllers/sync-volume/utils.go
```diff
@@ -160,3 +191,10 @@ func getConverter(vol *stosv1.Volume, obj storageos.Object, log logr.Logger) fun
 		return nil
 	}
 }
+
+func convertUint64ToInt64(val uint64) (int64, error) {
+	if val <= math.MaxInt64 {
+		return int64(val), nil
+	}
+	return 0, errors.New(fmt.Sprintf("cant typecast to int64: value is too large: %v", val))
```
mhmxs, line 199: `fmt.Errorf()` instead?

internal/controllers/sync-volume/utils.go
```diff
@@ -146,10 +159,28 @@ func getConverter(vol *stosv1.Volume, obj storageos.Object, log logr.Logger) fun
 				}
 
 				if replica.Health == stosv1.ReplicaHealthSyncing {
+					bytesRemaining, err := convertUint64ToInt64(stosReplica.SyncProgress.BytesRemaining)
+					if err != nil {
+						log.Info("failed to type cast sync bytes remaining")
```
mhmxs, line 164: Could you please give context to the error instead of info logging?

mhmxs (approved):

PR #112, opened 2022-01-18
Ingest new hostname and attachedOnHost fields from api

Ricardo-Osorio:
Ingest the new fields "hostname" and "attachedOnHost" that
were recently added to the openapi specs with version 2.6.0.
As part of this change, gomod was updated with the latest
version of the repo where this spec is defined.

Update gitignore to include vscode dir


internal/controllers/sync-volume/utils.go
```diff
@@ -134,6 +134,9 @@ func getConverter(vol *stosv1.Volume, obj storageos.Object, log logr.Logger) fun
 			}
 		}
 
+		fmt.Printf("master hostname %s\n", stosVol.Master.Hostname)
```
mhmxs, line 137: Do we need these lines in prod? If yes could you please convert it to logging?

Ricardo-Osorio: Not at all, those were using for debugging

mhmxs (approved):

PR #114, opened 2022-01-24
Use node id as label key instead of value

mhmxs:
This is the easiest way to find volume nodes via label selector.


Mojachieee (approved):
LGTM

PR #122, opened 2022-02-09
Update go-api dependency to use most recent version of the repo

Ricardo-Osorio:
Also had to change a few places where we had direct dependencies to auto generated object types (volume, node, etc) which had been manually edited.
This can certainly be improved in the future.


internal/pkg/storageos/node.go
```diff
@@ -44,11 +46,20 @@ func (c *Client) NodeObjects(ctx context.Context) (map[client.ObjectKey]Object,
 
 	nodes, resp, err := c.api.ListNodes(ctx)
 	if err != nil {
-		return nil, observeErr(stosapiv2.MapAPIError(err, resp))
+		return nil, observeErr(stosapierrors.MapAPIError(err, resp))
 	}
 	objects := make(map[client.ObjectKey]Object)
 	for _, node := range nodes {
-		objects[client.ObjectKey{Name: node.GetName()}] = node
+		obj := model.StosObject{
```
Mojachieee, line 53: nit: `nodeWrapper` instead of `obj`

internal/controllers/syncvolume_e2e_test.go
```diff
@@ -29,15 +31,24 @@ var _ = Describe("SyncVolume Controller", func() {
 			By("By creating a new volume")
 			stosVol := api.RandomStorageOSVolume()
 
-			err := api.AddVolume(*stosVol)
+			objWrapper := model.StosObject{
```
Mojachieee, line 34: nit: `volWrapper` instead of `objWrapper` is a bit more descriptive

internal/pkg/storageos/client.go
```diff
@@ -25,40 +26,35 @@ const (
 )
 
 //go:generate mockgen -build_flags=--mod=vendor -destination=mocks/mock_control_plane.go -package=mocks . ControlPlane
-//go:generate mockgen -build_flags=--mod=vendor -destination=mocks/mock_identifier.go -package=mocks . Identifier
```
Mojachieee, line 28: Out of interest why don't we need this mock anymore?

Mojachieee (approved):

internal/pkg/storageos/client.go
Ricardo-Osorio: That interface (a few line below in that same file) was merged into another interface.
It had a mere 3 method and didn't add anything other than increased complexity - we ended up having another mock obj.

internal/controllers/syncvolume_e2e_test.go
Ricardo-Osorio: Done

PR #127, opened 2022-03-22
Release of v1.2.7

mhmxs:


mhmxs:
I can't mere empty commit, so this is my bad. I would delete this PR and tag commit on master :( sorry

PR #130, opened 2022-04-05
Fix StorageOS volume hash calculation and status update

mhmxs:
In some cases after we patched volume object via `CreateOrPatch` it loads status from Kubernetes back, so next status update does nothung. To avoid this situation the new way applies status changes after `CreateOrPatch` so we always persist the latest state.

Meanwhile i was not happy to keep entire json objects in memory as volume hash, so i implemented a proper hash function.


internal/controllers/sync-volume/utils.go
```diff
@@ -41,7 +42,27 @@ import (
...
+		return nil, errors.New("type cast error")
+	}
+
+	raw, err := json.Marshal(stosVol)
+	if err != nil {
+		return nil, err
+	}
+
+	hash := []byte{}
+	for _, s := range [][]byte{[]byte(stosVol.Id), []byte(stosVol.CreatedAt.String()), raw} {
```
Mojachieee, line 56: What's the benefit of hashing the `id`, the `createdAt` and the entire `raw` json string?

Doesn't the entire json string contain the `id` and `createdAt` anyway?

mhmxs: The reason is we convert undefined number of any characters to ~19 digits. Because the output is slimmer than the input, we have some chance for collision. So this should power the output space and drastically decreases chance of collision.

Mojachieee (approved):

PR #136, opened 2022-06-03
Tune leader election

mhmxs:
Increase timeout of leader election.
Removal of legacy multi-lock to lease lock only.
Command-line flag to set renewal seconds.


main.go
```diff
@@ -234,14 +237,21 @@ func main() {
...
-		LeaderElection:          enableLeaderElection,
-		LeaderElectionID:        "storageos-api-manager-leader",
-		LeaderElectionNamespace: namespace,
+		Scheme:                     scheme,
+		MetricsBindAddress:         metricsAddr,
+		Port:                       9443,
+		LeaderElection:             enableLeaderElection,
+		LeaderElectionID:           "storageos-api-manager-leader",
+		LeaderElectionNamespace:    namespace,
+		LeaderElectionResourceLock: "leases",
```
Mojachieee, line 251: Same comment as before:
nit: I believe you can use the constant k8s.io/client-go/tools/leaderelection/resourcelock. LeasesResourceLock

Mojachieee (approved):

PR #144, opened 2022-07-28
CRD and controller logic for Pools

nolancon:
Introduces a new CRD `pool` and a controller for it. 

Key files to look at:
```
api/v1/pool_types.go
internal/controllers/pool/controller.go
internal/controllers/pool_e2e_test.go
internal/pkg/storageos/pool.go
```

The spec is defined here, anything listed below is not implemented yet though:
https://ondat.atlassian.net/browse/CTRL-240

Some things still need to be done:
 - Update status of CR - CTRL-291
 - Users should be able to provide parameters for StorageClass - CTRL-292
 - Finalizer should stop pool being deleted if volumes exist in pool - CTRL-293 

Other changes:
 - replaces uses of `ioutil` with `os` or `io` (as required by our linter) - it causes no functional changes


api/v1/pool_types.go
```diff
@@ -0,0 +1,71 @@
...
+SOFTWARE.
+*/
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NodeDriveMap maps which drives on which nodes are in the pool.
```
Ricardo-Osorio, line 31: `// NodeDriveMap maps which drives on which nodes are in the pool.`
I don't get what maps to what
Do you mean `// Maps nodes to storage devices`?

config/crd/bases/api.storageos.com_pools.yaml
```diff
@@ -0,0 +1,61 @@
...
+                  items:
+                    type: string
+                  type: array
+                description: NodeDriveMap maps which drives on which nodes are in
+                  the pool. A node is only in the pool if it has at least one drive
+                  that is in the pool
+                type: object
+            type: object
+          status:
+            description: PoolStatus defines the observed state of Pool
```
Ricardo-Osorio, line 49: I guess this confirms "status" is meant to be "state"?
Unless k8s makes this a pattern and should be named that way...
Let me know the reasoning behind it 🙏

internal/controllers/pool/controller.go
```diff
@@ -0,0 +1,210 @@
...
+			// - Add node ID to pool status
+			// - Check drives listed in NodeDriveMap, update status.DrivesList with
+			//   drives that the pool consists of.
+
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Pool exists on C2, update it
```
Ricardo-Osorio, line 130: We just called the GetPool so we could check if we need to update it instead of just sending the request through everytime

Ricardo-Osorio: Another thing, the UpdatePool is internally doing another GetPool? If possible it would be better to only do a single request to the API here
Maybe create a different UpdatePool that takes the object (struct from the api) when you have it already, which is this case

config/crd/bases/api.storageos.com_pools.yaml
Mojachieee: Status is a k8s thing. Each object has Metadata (name, namespace, labels, etc), a Spec (what the CR is meant to be) and a Status (what the CR currently is). 

In our case we're going to use the status to give a more user-friendly way of seeing which drives are in the pool.

api/v1/pool_types.go
```diff
@@ -0,0 +1,71 @@
...
+
+// PoolSpec defines the desired state of Pool
+type PoolSpec struct {
+	// NodeDriveMap maps which drives on which nodes are in the pool.
+	// A node is only in the pool if it has at least one drive that is in the pool
+	NodeDriveMap NodeDriveMap `json:"nodeDriveMap,omitempty"`
+}
+
+// PoolStatus defines the observed state of Pool
+type PoolStatus struct {
```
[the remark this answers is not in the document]
Mojachieee: Answered below

internal/controllers/pool/controller.go
Mojachieee: Both should be fixed now. 

UpdatePool no longer calls `GetPool` and we no longer update things unless we need to

api/v1/pool_types.go
Mojachieee: I've changed this comment, let me know if it makes sense now

Ricardo-Osorio (dismissed):

Ricardo-Osorio (approved):
LGTM

PR #146, opened 2022-08-24
Remove hardcoded values for spec inputs

nolancon:


Mojachieee (commented):
I left 2 minor comments, but apart from that it looks good

Ricardo-Osorio (approved):

PR #147, opened 2022-08-24
Fix broken release action - automatically create github release

Mojachieee:
Fixes an issue where the release action was always skipped - due to it both expecting the ref to be a branch and a tag.
Also, makes it so a github release is automatically created once the pipeline is finished


.github/workflows/release-image.yml
```diff
@@ -44,8 +43,13 @@ jobs:
         run: |
           docker tag ${{ env.OPERATOR_IMAGE }} ${{ env.RH_OPERATOR_IMAGE }}
           make operator-image-push OPERATOR_IMAGE=${{ env.RH_OPERATOR_IMAGE }}
+      - name: Create release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            storageos-operator.yaml
```
Ricardo-Osorio, line 50: Are these the right names given we are in the context of the api-manager repo?

Mojachieee: Nope, I was meant to delete those lines 😓

Mojachieee: Should be fixed now

Ricardo-Osorio (approved):

PR #149, opened 2022-08-26
Add finalizer to pool

nolancon:
This logic probably needs a kubecover test - it's difficult to test it properly with the existing kubebuilder testsuite.


Mojachieee (commented):
Looks good - pending kubecover tests

internal/controllers/pool/controller.go
```diff
@@ -113,15 +118,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-			if strings.Contains(err.Error(), "cannot delete pool: volumes exist in pool") {
+			if err == storageos.ErrPoolNotFound {
+				// nothing to do here
+			} else if strings.Contains(err.Error(), "cannot delete pool: volumes exist in pool") {
 				// Pool on C2 is being used by a volume and cannot be deleted at this time.
 				r.recorder.Event(pool, "Warning", "Cannot Delete Pool", "This pool is in use by a volume.")
-				// Reset deletion timestamp and return no error as we do not want to re-reconcile.
-				pool.ObjectMeta.DeletionTimestamp.Reset()
-				return ctrl.Result{}, r.Client.Update(ctx, pool)
+				r.Client.Update(ctx, pool)
```
nolancon, line 126: I think this makes most sense here:
```
if err := r.Client.Update(ctx, pool); err != nil {
  // Requeue so we can try again in the future
  return ctrl.Result{RequeueAfter: 20 * time.Second}, err
}
return ctrl.Result{}, nil

nolancon, line 126: I don't think we need this update call now if we're no longer resetting the timestamp?

Mojachieee: I didn't know if we needed to call update to add the event

PR #155, opened 2022-10-14
Webhook license check

nolancon:


controllers/pvc-mutator/controller.go
```diff
@@ -66,3 +76,38 @@ func (c *Controller) Handle(ctx context.Context, req admission.Request) admissio
...
+	var storageRequested int64
+	var ok bool
+	if storageRequested, ok = pvc.Spec.Resources.Requests.Storage().AsInt64(); !ok {
+		if storageRequested, ok = pvc.Spec.Resources.Requests.Storage().ToDec().AsInt64(); !ok {
+			return fmt.Errorf("failed to get int64 representatoin of requested storage")
+		}
+	}
+
+	// Get licence from control plane.
+	licence, err := c.licence.GetLicence(ctx)
```
mhmxs, line 103: Does it make sense to cache the result for a time period? I mean usually licence doesn't change in a minute bases.

mhmxs (approved):

PR #157, opened 2022-11-30
Updates docker hub creds

anas-domesticus:


Ricardo-Osorio:
I have updated the whole actions steps in another commit, if you don't mind I will just add the updated creds there so there's no need for extra work rebasing or merging

PR #158, opened 2023-01-16
Fix memory leak of volume and node cache

mhmxs:
This change fixes a theoretical memory leak. I write theoretical because the customer has to create and delete tons of volumes to feel the pain.


Mojachieee (approved):
Nice catch!

PR #161, opened 2023-02-06
Refresh api token only if elected

mhmxs:


main.go
```diff
@@ -241,14 +240,15 @@ func main() {
...
+	leaderRetryDuration := renewDeadline / 2
 
 	// Only attempt to grab leader lock once we have an API connection.
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                     scheme,
 		MetricsBindAddress:         metricsAddr,
 		Port:                       9443,
 		LeaderElection:             enableLeaderElection,
-		LeaderElectionID:           "storageos-api-manager-leader",
+		LeaderElectionID:           "storageos-api-manager-leader-asdekfww",
```
Ricardo-Osorio, line 251: "asdekfww" if we want to use a different name we could use something that made sense 😄

Ricardo-Osorio (approved):

main.go
mhmxs: This is a pattern i have seen, to give a unique ID for the lease. I know we don't necessary have to change this, but why not :)

